PR #22, opened 2019-05-01
Feat/test local

marty331:


subhub/tests/conftest.py
```diff
@@ -1,12 +1,52 @@
...
+    proc = psutil.Process(pid=ddb_process.pid)
+    child_procs = proc.children(recursive=True)
+    for p in [proc] + child_procs:
+        os.kill(p.pid, signal.SIGTERM)
+    ddb_process.wait()
+
 
 @pytest.fixture(scope="module")
 def app():
+    print(f'test app')
```
sidler-mozilla: 'f' for formatting not required for this string as there is no {} interpolation

marty331: agree

PR #30, opened 2019-05-07
SS-134 Add SecurityDefinitions to subhub_api.yaml for validation 

marty331:
- validate API tokens for multiple systems
-  allow for multiple API token types
-  test all methods in  auth_validation.py

### DIFF
--- a/subhub/tests/unit/test_auth_validation.py
+++ b/subhub/tests/unit/test_auth_validation.py
@@ -1,0 +1,38 @@
+import json
+
+from subhub import auth_validation
+from subhub.cfg import CFG
+
+
+def test_payment_auth():
+    api_token = 'sh_test_testkey'
+    payments_auth = auth_validation.payment_auth(api_token, required_scopes=None)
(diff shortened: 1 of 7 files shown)


subhub/tests/unit/test_auth_validation.py
```diff
@@ -0,0 +1,38 @@
+import json
+
+from subhub import auth_validation
+from subhub.cfg import CFG
+
+
+def test_payment_auth():
+    api_token = 'sh_test_testkey'
+    payments_auth = auth_validation.payment_auth(api_token, required_scopes=None)
+    assert payments_auth["token"] in CFG.PAYMENT_API_KEY
```
sidler-mozilla: you should really be consistent on quoting; single quotes for 'sh_test_testkey' and double for "token"; really applies all through these changes.

bbangert: This is where it'd be nice to just use black and call it a day. It's a pain to remember which one to use (I switch code-bases a lot, which have different styles), having an auto-formatter is sanity-saving.

PR #32, opened 2019-05-08
feat: refactor helper functions, add more thorough error handling

bbangert:
Refactors the helper functions to a customer.py module for individual
testing as well as removing any dependency on Flask state.

Added type hints to all customer helper functions and most of the API
methods.

Added error handling for Stripe errors to Flask itself for catching.
This removes duplicate handling inside functions and provides a uniform
place to translate errors as appropriate.


marty331:
The create_customer method is looking for a list of customers by email (smart) but it does not check if the customer has the metadata userid that matches the sent user_id, therefore it matches to a customer that does not have the correct metadata.  

The correction for this could cause an AttributeError to be thrown as it's looking for metadata that possibly doesn't exist, so the exceptions.py should include and handle.

PR #46, opened 2019-05-13
Cancel subscription month end

marty331:


subhub/api/payments.py
```diff
@@ -72,23 +74,23 @@ def cancel_subscription(uid, sub_id) -> FlaskResponse:
...
                 return {"message": e}, 400
-            if "No such subscription:" in tocancel:
-                return {"message": "Invalid subscription."}, 404
-            if tocancel["status"] in ["active", "trialing"]:
-                tocancel.delete()
-                return {"message": "Subscription cancellation successful"}, 201
-            else:
-                return {"message": "Error cancelling subscription"}, 400
     else:
         return {"message": "Subscription not available."}, 400
```
bbangert: Is this going to end up returning 'Subscription not available' if the subscription was already cancelled?

subhub/tests/unit/test_payment_calls.py
```diff
@@ -136,6 +136,38 @@ def test_create_subscription_with_valid_data():
     g.subhub_account.remove_from_db("validcustomer")
 
 
+def test_subscribe_customer_existing(create_customer_for_processing):
```
bbangert: Probably want to add a test to show what happens when you cancel a subscription that was already cancelled. I would expect that in that case, the command would still go through ok, even though nothing needs to be done. If an error is thrown, then it should indicate that it was already cancelled.

subhub/api/payments.py
```diff
@@ -101,7 +103,7 @@ def subscription_status(uid) -> FlaskResponse:
     subscriptions = stripe.Subscription.list(
         customer=items.custId, limit=100, status="all"
     )
-    if subscriptions is None:
+    if len(subscriptions) == 0:
```
bbangert: I'm used to seeing this check done as:
```python
if not subscriptions:
```

Was there some other reason not to rely on typical [Python truth value testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)?

subhub/tests/unit/test_payment_calls.py
```diff
@@ -136,6 +136,38 @@ def test_create_subscription_with_valid_data():
...
+    WHEN provided a customer and plan
+    THEN validate subscription is created
+    """
+    uid = uuid.uuid4()
+    subscription, code = payments.subscribe_to_plan(
+        "validcustomer",
+        {
+            "pmt_token": "tok_visa",
+            "plan_id": "plan_EtMcOlFMNWW4nd",
+            "email": "valid@{}customer.com".format(uid),
```
sidler-mozilla: why not an f-string?

subhub/tests/unit/test_payment_calls.py
```diff
@@ -136,6 +136,38 @@ def test_create_subscription_with_valid_data():
...
+            "orig_system": "Test_system",
+        },
+    )
+    subscription2, code2 = payments.subscribe_to_plan(
+        "validcustomer",
+        {
+            "pmt_token": "tok_visa",
+            "plan_id": "plan_EtMcOlFMNWW4nd",
+            "email": "valid@{}customer.com".format(uid),
```
sidler-mozilla: f-string

subhub/tests/unit/test_payment_calls.py
```diff
@@ -203,6 +235,17 @@ def test_cancel_subscription_with_valid_data(app, create_subscription_for_proces
     g.subhub_account.remove_from_db("process_test")
 
 
+def test_cancel_subscription_with_invalid_data(app, create_subscription_for_processing):
+    (subscription, code) = create_subscription_for_processing
+    (cancelled, code) = payments.cancel_subscription(
+        "process_test", subscription["subscriptions"][0]["subscription_id"]+"a"
```
sidler-mozilla: single char "a" is somewhat ambiguous; why not +"invalid" to make it more explicit?

subhub/api/payments.py
marty331: @bbangert yes it would return 'Subscription not available'.  Is that an issue?  Would changing the message be better or do you want to re-cancel?

bbangert: In the event of a time-out, FxA would not see that it was cancelled successfully and retry. So a repeat call should still work, it just won't need to cancel.

subhub/api/payments.py
marty331: Oversight, I corrected it.

subhub/api/payments.py
```diff
@@ -46,14 +46,16 @@ def list_all_plans() -> FlaskListResponse:
     plans = stripe.Plan.list(limit=100)
     stripe_plans = []
     for p in plans:
```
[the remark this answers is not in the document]
marty331: good catch, changed.

subhub/api/payments.py
marty331: OK I changed to this functionality and validated that it will not throw an error.

bbangert: This PR hasn't been updated yet to change that, did you forget to push the change?

subhub/tests/unit/test_payment_calls.py
marty331: Done.

subhub/tests/unit/test_payment_calls.py
marty331: It is now.

subhub/tests/unit/test_payment_calls.py
marty331: Done.

subhub/tests/unit/test_payment_calls.py
marty331: Good suggestion.

subhub/api/payments.py
marty331: No, doing a final test and will then push.

subhub/api/payments.py
```diff
@@ -181,6 +191,12 @@ def create_update_data(customer) -> dict:
     :param customer:
     :return: return_data dict
     """
+    for subscription in customer["subscriptions"]["data"]:
+        if subscription["status"] == "incomplete":
+            invoice = stripe.Invoice.retrieve(subscription["latest_invoice"])
+            print(f"invoice {invoice}")
```
bbangert: Print statements shouldn't be included. To avoid having to include/exclude them during local dev, you could use log.debug statements instead.

Issue #43, opened 2019-05-13
Update README to indicate how to setup and develop locally


bbangert:
To run subhub locally for FxA, we need to know how to run it locally. The README or some other document should indicate the system dependencies, requirements, and steps needed for OSX/Linux to have a working subhub locally.

marty331:
This is resolved with: https://github.com/mozilla/subhub/pull/55

PR #50, opened 2019-05-15
Update README.md with some local dev help

lmorchard:
- Also tweak local DBD host to accept env var

Issue #43

### DIFF
--- a/subhub/app.py
+++ b/subhub/app.py
@@ -16,7 +16,7 @@ def create_app(config=None):
     if not CFG("AWS_EXECUTION_ENV", None):
         options = {"swagger_ui": True}
         region = "localhost"
-        host = "http://localhost:8000"
+        host = CFG("DBD_HOST", "http://localhost:8000")
         stripe.api_key = CFG.STRIPE_API_KEY
     else:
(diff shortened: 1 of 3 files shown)


lmorchard:
Not sure if this helps break the ice on the issue, but figured I'd try a PR with some notes I tried jotting down while getting myself going. I am most likely missing things, so feedback is appreciated

subhub/app.py
```diff
@@ -16,7 +16,7 @@ def create_app(config=None):
     if not CFG("AWS_EXECUTION_ENV", None):
         options = {"swagger_ui": True}
         region = "localhost"
-        host = "http://localhost:8000"
+        host = CFG("DBD_HOST", "http://localhost:8000")
```
lmorchard: Oh yeah, and I had to tweak this port because it turns out all the usual ports on my machine are crowded thanks to the FxA stack consisting of 20 microservices or so already.

marty331: @lmorchard just FYI we are reviewing some options on our end for making this easier.

lmorchard:
Going to go ahead and close this since it looks like PR #55 is doing quite a bit more

PR #53, opened 2019-05-16
refactored the exceptions classes and added 100% test coverage

sidler-mozilla:


subhub/exceptions.py
```diff
@@ -1,52 +1,58 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+
 class SubHubError(Exception):
     """Base SubHub Exception"""
 
     status_code = 500
 
+    def __init__(self, message, status_code, payload):
```
bbangert: How come this one doesn't call super().__init__ for the Exceptions base init?

sidler-mozilla: it should; refactoring

PR #55, opened 2019-05-17
add task to run locally

sidler-mozilla:
This is a big commit with lots of changes:
- README.md
multiple entries with more detail for all of the doit tasks available.  This can probably still use more, but this is a big improvement for now.
- dodo.py
The setup task is removed in favor of npm which operates at the REPOROOT rather than on each service/<name>.  Task for setting up venv has been added and task for starting and stopping dyanlite local db.  These tasks are in support of the new local task for running the subhub flask app locally.
- package.json files
These files were moved to the REPOROOT of the git repo.
- subhub/cfg.py
Several properties were added for use with the CFG.  
- subhub/tests/conftest.py
The pathing was adjusted to match node_modules in the REPOROOT.  Added use of DYNALITE_PORT.
- subhub/tests/test_cfg.py
Added tests for new properties to bring coverage to 98%.  It doesn't seem possible to get to 100% due to the nature of the code.


subhub/app.py
```diff
@@ -13,7 +13,7 @@
 
 
 def create_app(config=None):
-    if not CFG("AWS_EXECUTION_ENV", None):
+    if not CFG.AWS_EXECUTION_ENV:
         options = {"swagger_ui": True}
         region = "localhost"
         host = "http://localhost:8000"
```
lmorchard: FWIW, looks like `DYNALITE_PORT` should get used here too. Had to make that tweak locally to get this working.

dodo.py
```diff
@@ -138,6 +139,100 @@ def task_noroot():
...
+    '''
+    cmd = f'{DYNALITE} --port {CFG.DYNALITE_PORT}'
+    msg = f'dyanlite server started on {CFG.DYNALITE_PORT} logging to {CFG.DYNALITE_FILE}'
+    def running():
+        '''
+        return pid of running dyanlite
+        '''
+        pid = None
+        try:
+            pid = call('lsof -i:8000 -t')[1].strip()
```
lmorchard: Another spot for `CFG.DYNALITE_PORT`

subhub/app.py
sidler-mozilla: oh good catch!

dodo.py
sidler-mozilla: like bugs, needing a squashing

dodo.py
```diff
@@ -138,6 +139,100 @@ def task_noroot():
...
+    ])
+    return {
+        'task_dep': [
+            'noroot',
+            'venv',
+            'dynalite:start',
+        ],
+        'actions': [
+            f'{python3} -m setup develop',
+            f'{aws} {python3} subhub/app.py',
```
lmorchard: This might be an Ubuntu on WSL thing, but to get this working I also had to tweak this like so:
```python
f'{aws} {python3} -m subhub.app',
```

PR #71, opened 2019-05-24
Feat/remove payment

marty331:
SS-96 Remove payment sources
- If all subscriptions are cancelled or set to..
  be cancelled, then remove all payment sources
- Test all new code
- Implemented mockito to test Stripe calls..
  offline for true unit testing

### DIFF
--- a/subhub/api/payments.py
+++ b/subhub/api/payments.py
@@ -12,9 +12,14 @@ from flask import g
 
 from subhub.api.types import JsonDict, FlaskResponse, FlaskListResponse
 from subhub.customer import existing_or_new_customer, has_existing_plan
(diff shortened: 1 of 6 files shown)


subhub/api/payments.py
```diff
@@ -137,39 +178,51 @@ def create_return_data(subscriptions) -> JsonDict:
     return_data = dict()
     return_data["subscriptions"] = []
     for subscription in subscriptions["data"]:
+        # logger.info(f"subscription {subscription}")
```
jackiemunroe: Should these logging lines be removed?

marty331: Yes, I'll remove and update.

PR #81, opened 2019-06-01
Stripe error message

jackiemunroe:
Updated responses to include Stripe's error messaging. Added tests to increase code coverage.

### DIFF
--- a/subhub/api/payments.py
+++ b/subhub/api/payments.py
@@ -1,13 +1,6 @@
 import logging
 from datetime import datetime
 import stripe
-from stripe.error import (
-    InvalidRequestError,
-    APIConnectionError,
-    APIError,
-    AuthenticationError,
-    CardError,
-)
 from flask import g
 
 from subhub.api.types import JsonDict, FlaskResponse, FlaskListResponse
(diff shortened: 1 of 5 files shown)


subhub/api/payments.py
```diff
@@ -33,19 +27,14 @@ def subscribe_to_plan(uid, data) -> FlaskResponse:
...
-    try:
+        existing_plan = has_existing_plan(customer, plan_id=data["plan_id"])
+        if existing_plan:
+            return {"message": "User already subscribed."}, 409
         stripe.Subscription.create(
             customer=customer.id, items=[{"plan": data["plan_id"]}]
         )
-    except InvalidRequestError as e:
-        return {"message": f"Unable to subscribe: {e}"}, 400
+    except StripeError as e:
```
bbangert: Why is this catching [every single Stripe error ](https://github.com/stripe/stripe-python/blob/master/stripe/error.py)(including auth and rate limit issues in SubHub) and returning them as client errors?

bbangert:
I'm a little confused why Stripe error handling is creeping back into the `payments.py` file, when the Flask error catching can blanket catch all of them at the view level. When I refactored to use them, I removed all the Stripe error catching from the first `payments.py` method in favor of letting the Flask error handler blanket catch it.

Here are the docs on the Flask error handling that is in use:
http://flask.pocoo.org/docs/1.0/errorhandling/#error-handlers

Any exception from a view in `payments.py` triggers the appropriate error handler defined in `app.py`. Is this error handling not working correctly for some reason?

bbangert:
One other note, there currently are no tests verifying the app is configured correctly with its error handling, as that would require creating the app. This is done in one place, so if you wanted to add some integration tests that verify the app works with the error handling that should be pretty straight forward. It could go in https://github.com/mozilla/subhub/blob/master/subhub/tests/unit/test_subhub.py, and after the app is created [Flask includes some great testing docs](http://flask.pocoo.org/docs/1.0/testing/).

This way you can test the app at the API layer to see that the API routing works correctly all the way through to the error catchers.

subhub/api/payments.py
jackiemunroe: @bbangert I updated it to catch all Stripe errors, since the request was to show the Stripe error messages. I left it as a 400 error, since that was how we were handling it before. I do not mind setting it back to only catching the InvalidRequestError.

jackiemunroe:
@bbangert As a note: All of the exception handling that was updated to catching StripeError, were already catching errors from Stripe, I just updated them to catch their base class because in some of the cases the line was fairly long and about half of the StripeErrors defined by Stripe. So, StripeError handling was still already happening within payments.py.

I do not mind updating how we are handling errors, I just want to clarify that the purpose of this task was to update our responses to send the appropriate StripeError. 
It is entirely possible that I misinterpreted the task as defined:
https://jira.mozilla.com/browse/SS-226

Please let me know how to proceed

jackiemunroe:
After offline discussions with @bbangert , I will be updating to remove StripeError handling from within payments.py and update the response formatting for StripeErros in app.py as well as adding tests for this functionality.

PR #90, opened 2019-06-05
Fix cancellation bug

jackiemunroe:
Updated cancel_subscription method to ensure that in the case of multiple subscriptions, that the loop would reach past the first instance when searching for the subscription to cancel

Added tests covering the multiple subscription cases - cancelling first and cancelling second

### DIFF
--- a/subhub/tests/unit/test_payment_calls.py
+++ b/subhub/tests/unit/test_payment_calls.py
@@ -246,6 +246,90 @@ def test_cancel_subscription_with_valid_data(app, create_subscription_for_proces
     g.subhub_account.remove_from_db("process_test")
 
 
+def test_cancel_subscription_with_valid_data_multiple_subscriptions_remove_first():
+    """
+    GIVEN a user with multiple subscriptions
+    WHEN the first subscription is cancelled
+    THEN the specified subscription is cancelled
+    """
+    uid = uuid.uuid4()
+    subscription1, code1 = payments.subscribe_to_plan(
+        "valid_customer",
+        {
+            "pmt_token": "tok_visa",
+            "plan_id": "plan_EtMcOlFMNWW4nd",
+            "email": f"valid@{uid}customer.com",
+            "orig_system": "Test_system",
+        },
+    )
+    subscription2, code2 = payments.subscribe_to_plan(
+        "valid_customer",
+        {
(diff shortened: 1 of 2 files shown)


subhub/tests/unit/test_payment_calls.py
```diff
@@ -246,6 +246,82 @@ def test_cancel_subscription_with_valid_data(app, create_subscription_for_proces
...
+    (cancelled, code) = payments.cancel_subscription(
+        "valid_customer", subscription1["subscriptions"][0]["subscription_id"]
+    )
+    assert cancelled["message"] == "Subscription cancellation successful"
+    assert 201 == code
+
+    g.subhub_account.remove_from_db("valid_customer")
+    (cancelled, code) = payments.cancel_subscription(
+        "valid_customer", subscription2["subscriptions"][0]["subscription_id"]
+    )
```
bbangert: Would it be useful to verify what the code returned here is?

subhub/tests/unit/test_payment_calls.py
```diff
@@ -246,6 +246,82 @@ def test_cancel_subscription_with_valid_data(app, create_subscription_for_proces
...
+    assert cancelled["message"] == "Subscription cancellation successful"
+    assert 201 == code
+
+    g.subhub_account.remove_from_db("valid_customer")
+    (cancelled, code) = payments.cancel_subscription(
+        "valid_customer", subscription1["subscriptions"][0]["subscription_id"]
+    )
```
bbangert: Same here, if you cancel it twice, what's the expected behavior?

jackiemunroe: It's how I saw we were cleaning up the tests data. So the purpose here is to clean up the other instance of test data.

subhub/tests/unit/test_payment_calls.py
jackiemunroe: Yes it would be useful to verify that it also had been removed. I will add the check.

subhub/tests/unit/test_payment_calls.py
jackiemunroe: I have added a comment for clarity and added additional checks to ensure that the clean-up is successful

PR #99, opened 2019-06-12
Implement SNS Topics

marty331:
- Push Firefox events to SNS topic for subscriber
- Refactor SubscriptionCreated per FxA requirements
- Refactor SubscriptionDeleted per FxA requirements


services/fxa/serverless.yml
```diff
@@ -150,6 +155,9 @@ functions:
           method: ANY
           path: '{proxy+}'
           cors: true
+      - sns:
+          topicName: fxa-data
```
sidler-mozilla: any reason why this isnt just `fxa-event-data` to match more closely the Display Name?

subhub/api/webhooks/stripe/customer.py
```diff
@@ -63,53 +65,58 @@ def run(self):
...
-        routes = [StaticRoutes.SALESFORCE_ROUTE, StaticRoutes.FIREFOX_ROUTE]
-        self.send_to_routes(routes, json.dumps(data))
+        user_id = self.payload.data.object.metadata.get("userid", None)
+        updated_customer = ""
+        if not user_id:
+            customer_id = self.payload.object.customer
+            updated_customer = stripe.Customer.retrieve(customer_id)
+            user_id = updated_customer.metadata.get("userid")
+        sub_status = False
+        if self.payload.data.object.status in ["active", "trialing"]:
```
sidler-mozilla:
```
sub_status = self.payload.data.object.status in ("active", "trialing")
```
???

subhub/api/webhooks/stripe/customer.py
```diff
@@ -63,53 +65,58 @@ def run(self):
...
+                productName=self.payload.data.object.plan.nickname,
+                eventId=self.payload.id,
+                event_id=self.payload.id,
+                eventCreatedAt=self.payload.created,
+                messageCreatedAt=int(time.time()),
+            )
+            routes = [StaticRoutes.FIREFOX_ROUTE]
+            self.send_to_routes(routes, json.dumps(sfd))
+        else:
+            logger.exception(f"userid is None for customer {updated_customer}")
```
sidler-mozilla: else logger exception again, this is correct here?

subhub/api/webhooks/stripe/customer.py
```diff
@@ -63,53 +65,58 @@ def run(self):
...
-        routes = [StaticRoutes.SALESFORCE_ROUTE, StaticRoutes.FIREFOX_ROUTE]
-        self.send_to_routes(routes, json.dumps(data))
+        user_id = self.payload.data.object.metadata.get("userid", None)
+        updated_customer = ""
+        if not user_id:
+            customer_id = self.payload.object.customer
+            updated_customer = stripe.Customer.retrieve(customer_id)
+            user_id = updated_customer.metadata.get("userid")
+        sub_status = False
+        if self.payload.data.object.status in ["active", "trialing"]:
```
sidler-mozilla: same as above?

subhub/cfg.py
```diff
@@ -336,32 +336,42 @@ def PAYMENT_API_KEY(self):
...
     @property
-    def SUPPORT_API_KEY(self):
+    def TOPIC_ARN(self):
         """
-        support api key
+        topic arn for sns
+        :return:
         """
-        return self("SUPPORT_API_KEY", "fake_support_api_key")
+        return self("TOPIC_ARN", "fake_arn_key")
```
sidler-mozilla: this doesnt follow the pattern laid out before

`SUPPORT_API_KEY` -> `fake_support_api_key`
so it would be
`TOPIC_ARN` -> `fake_topic_arn`

subhub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -1,11 +1,17 @@
+import time
+import json
+
 import mockito
 import requests
 import boto3
+from botocore.exceptions import ClientError
```
sidler-mozilla: group the from imports separate from the regular imports

subhub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -101,64 +107,55 @@ def test_stripe_webhook_customer_source_expiring(mocker):
...
-        "plan_amount": 500,
-        "plan_currency": "usd",
-        "plan_interval": "month",
-        "status": "canceled",
-        "trial_start": None,
-        "trial_end": None,
-        "tax_percent": None,
-        "application_fee_percent": None,
-        "user_id": None,
+        "uid": "tester123",
```
sidler-mozilla: food for thought the tests could use dict() + keyword instead of {} and 'quoted-strings' just like the app code does now...

subhub/tests/unit/stripe/utils.py
```diff
@@ -22,6 +22,11 @@ def send_message(QueueUrl={}, MessageBody={}):
         return {"ResponseMetadata": {"HTTPStatusCode": 200}}
 
 
+class MockSnsClient:
+    def publish(Message: dict = {}, MessageStructure: str = "json", TopicArn: str = ""):
```
sidler-mozilla: bug: don't assign keyword args mutable objects like dicts
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

subhub/api/webhooks/stripe/customer.py
marty331: Great catch, implementing this.

subhub/api/webhooks/stripe/customer.py
marty331: Great catch, implementing this.

subhub/api/webhooks/stripe/customer.py
marty331: Changed to raise ClientError so this will be logged.  Chances of this happening are extremely slim as the userid is required to setup an account in the first place.

services/fxa/serverless.yml
marty331: changed to ${self:provider.stage}-fxa-data as we are defining the SNS Topics as such and this is how FxA is subscribing.

subhub/cfg.py
marty331: Changed to TOPIC_ARN_KEY -> fake_topic_arn_key

subhub/tests/unit/stripe/customer/test_stripe_customer.py
marty331: I had not heard of this separation guideline before but I implemented.

subhub/tests/unit/stripe/utils.py
marty331: Interesting, OK I changed this to  have None for Message and TopicArn, tests still pass.

subhub/tests/unit/stripe/customer/test_stripe_customer.py
marty331: I implemented this on this function, we can update the others in a future pass.

marty331:
> See what you think about all of those comments.

I think they were all well thought out and I amended the code.

subhub/api/webhooks/routes/firefox.py
```diff
@@ -14,24 +16,21 @@
...
+                Message=json.dumps({"default": json.dumps(self.payload)}),
+                MessageStructure="json",
+            )
+            logger.info(f"message {self.payload}")
+            if response["ResponseMetadata"]["HTTPStatusCode"] == 200:
+                logger.info(f"message sent to Firefox queue: {response}")
                 route_payload = json.loads(self.payload)
                 self.report_route(route_payload, "firefox")
+                return response
             else:
```
jackiemunroe: since the the `if` ends with a return, the `else` is unnecessary

subhub/api/webhooks/routes/firefox.py
```diff
@@ -11,24 +13,19 @@
...
-
-            if msg["ResponseMetadata"]["HTTPStatusCode"] == 200:
-                logger.info(f"message sent to Firefox queue: {msg}")
+            sns_client = boto3.client("sns", region_name=CFG.AWS_REGION)
+            response = sns_client.publish(
+                TopicArn=CFG.TOPIC_ARN_KEY,
+                Message=json.dumps({"default": json.dumps(self.payload)}),
+                MessageStructure="json",
+            )
+            logger.info(f"message {self.payload}")
```
jackiemunroe: Should this log string be updated?

marty331: Definitely and it should have been included in the last force push, I'll fix.

subhub/api/webhooks/stripe/customer.py
```diff
@@ -59,53 +62,54 @@ def run(self):
...
-            "plan_name": self.payload.data.object.plan.nickname,
-            "trial_period_days": self.payload.data.object.plan.trial_period_days,
-            "status": self.payload.data.object.status,
-            "canceled_at": self.payload.data.object.canceled_at,
-            "created": self.payload.data.object.created,
-        }
-        routes = [StaticRoutes.SALESFORCE_ROUTE, StaticRoutes.FIREFOX_ROUTE]
-        self.send_to_routes(routes, json.dumps(data))
+        user_id = self.payload.data.object.metadata.get("userid", None)
+        updated_customer = ""
```
sidler-mozilla: why are we setting this to the empty string?

marty331: Good catch, changed to self.payload.object.customer in the error log below.

PR #109, opened 2019-06-27
Add method to Delete Customer

marty331:
- Added delete_customer method to payments.py
  - Deletes customer from payment provider
  - Cancels any active subscriptions for customer
- Added tests for delete_customer method
- Added DELETE /customer/{uid} route

### DIFF
--- a/subhub/api/payments.py
+++ b/subhub/api/payments.py
@@ -140,6 +140,24 @@ def cancel_subscription(uid, sub_id) -> FlaskResponse:
     return {"message": "Subscription not available."}, 400
 
 
+def delete_customer(uid) -> FlaskResponse:
+    """
+    Delete an existing customer, cancel active subscriptions
+    and delete from payment provider
+    :param uid:
+    :return: Success of failure message for the deletion
+    """
+    subscription_user = g.subhub_account.get_user(uid)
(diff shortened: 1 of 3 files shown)


subhub/api/payments.py
```diff
@@ -140,6 +140,24 @@ def cancel_subscription(uid, sub_id) -> FlaskResponse:
...
+    :return: Success of failure message for the deletion
+    """
+    subscription_user = g.subhub_account.get_user(uid)
+    if not subscription_user:
+        return dict(message="Customer does not exist."), 404
+    deleted_payment_customer = stripe.Customer.delete(subscription_user.custId)
+    if deleted_payment_customer:
+        deleted_customer = g.subhub_account.remove_from_db(uid)
+        if deleted_customer:
+            return dict(message="Customer deleted successfully"), 201
```
sidler-mozilla: is `201 Created` the right code to return when deleting something?

sidler-mozilla: "according to the Googs, status code for successful DELETE _should_ be 200" -afrank

marty331: After looking into this I think 200 is probably the correct value, so I've changed the code to return a 200.

PR #117, opened 2019-07-05
added pytest to allow running tests in one file at a time

sidler-mozilla:


dodo.py
```diff
@@ -413,6 +413,25 @@ def task_test():
...
+            'name': filename,
+            'task_dep': [
+                'check',
+                'stripe',
+                'yarn',
+                'venv',
+                'dynalite:stop',
+            ],
+            'actions': [
+                f'{PYTHON3} -m pytest {filename} --disable-warnings -vxs',
```
bowlofstew: Do you all typically want JUnit style XML/HTML/record test reports to be generated, etc for the CI side of things?

sidler-mozilla: this is just for being able to run a single test file, without having to run the entire suite.  the task that is usually run is the `task_test` which uses `tox`. I am not against see what xml|html looks like but usually this will be printed to the term or in travis-ci, the logs.  Seems xml|html would be less than ideal in those scenarios.

marty331: Printing the results like this is exactly what I would like.

sidler-mozilla: xml|html will not look good in logs like this https://travis-ci.org/mozilla/subhub/builds/554829680 
So I would be interested in what we are talking about.

PR #121, opened 2019-07-09
Initial requirement extraction and dependency versioning

bowlofstew:
This pull request (PR) is an idea that @scottidler and I were discussing last week.  It replaces the `setup.py` application and setup dependencies with contents from associated, prefixed requirements files.  Secondly, this PR serves to version the various dependencies to avoid the issue we observed last week in which 2 dependencies were being silently upgraded and causing run-time errors.


bowlofstew:
Looking into the build failure due to tests failure now.

setup.py
```diff
@@ -7,66 +7,14 @@
...
-    'pytest-mock',
-    'flake8',
-    'mock',
-    'mockito',
-    'mockito',
-    'patch',
-    'pytest',
-    'pytest',
-    'pytest-cov',
-    'pytest-cov',
```
bowlofstew: Notice that we had multiple, duplicates within this file.  This change set when migrated generated an packaging error on `doit package` that would prevent this moving forward.

PR #122, opened 2019-07-10
fix nr env vars by passing them in

sidler-mozilla:


subhub/cfg.py
```diff
@@ -406,6 +406,34 @@ def SWAGGER_UI(self):
...
+        NEW_RELIC_ACCOUNT_ID
+        """
+        return self("NEW_RELIC_ACCOUNT_ID", 2239138)
+
+    @property
+    def NEW_RELIC_TRUSTED_ACCOUNT_ID(self):
+        """
+        NEW_RELIC_TRUSTED_ACCOUNT_ID
+        """
+        return self("NEW_RELIC_TRUSTED_ACCOUNT_ID", 2239138)
```
bowlofstew: Is this a reasonable default for the application?  I don't know what it actually is pointing at.

subhub/cfg.py
```diff
@@ -406,6 +406,34 @@ def SWAGGER_UI(self):
         """
         return self.APP_DEPENV in ("stage", "qa", "dev")
 
+    @property
+    def NEW_RELIC_ACCOUNT_ID(self):
+        """
+        NEW_RELIC_ACCOUNT_ID
+        """
+        return self("NEW_RELIC_ACCOUNT_ID", 2239138)
```
bowlofstew: below comment's question applies.

PR #125, opened 2019-07-12
added logger.error for every except in our code

sidler-mozilla:


subhub/customer.py
```diff
@@ -57,7 +58,9 @@ def create_customer(
     if not subhub_account.save_user(db_account):
         # Clean-up the Stripe customer record since we can't link it
         stripe.Customer.delete(customer.id)
-        raise IntermittentError("error saving db record")
+        e = IntermittentError("error saving db record")
+        logger.error("unable to save user or linke it", error=e)
```
bowlofstew: `linke`?

PR #126, opened 2019-07-12
Visual Profiling of Requests

bowlofstew:
This pull request (PR) provides visual profiling of incoming requests and payloads from a caller.  It was branched from [#124 ](https://github.com/mozilla/subhub/pull/124) so it would require that going back to master prior to this branch coming back.  This was done so that the annotations could be used for determination of memory as well.

### DIFF
--- a/subhub/app.py
+++ b/subhub/app.py
@@ -1,10 +1,12 @@
 import os
+import sys
 
 import connexion
 import stripe
 import stripe.error
 from flask import current_app, g, jsonify
 from flask_cors import CORS
+from flask import request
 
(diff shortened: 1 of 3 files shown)


subhub/app.py
```diff
@@ -106,9 +107,24 @@ def server_stripe_card_error(e):
 
     @app.app.before_request
     def before_request():
+        from flask import request
```
sidler-mozilla: let's move this to the top of the file with the rest of the Flask imports

bowlofstew: Cool, will do!

PR #130, opened 2019-07-16
Removed Profiling Annotations

bowlofstew:
This pull request (PR) removes the profiling annotations that I added for performance profiling and tracing in our applications.  I had not ran into this locally yet so I have removed them and will go thru a reproduction effort now and wrap it under a unit test then resolve.  Thanks for the report @lmorchard.


lmorchard:
FWIW, belated :+1: - this appears to resolve the issue locally for me. Still seeing the 500 in our dev stack, but not sure if this gets continually-deployed from master

PR #131, opened 2019-07-16
Initial remote & local performance tests

bowlofstew:
This pull request (PR) provides initial support for:
- the Locust load testing framework
- running Locust against local environments
- running Locust against remote environments
- supporting many payment_api_keys
- supporting many customers
- supports the customer workflow
I fully expect comments on the implementation and `doit` and welcome it still.  I just wanted to start getting this work out in front of others.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -52,6 +52,7 @@ def envs(sep=' ', **kwargs):
         NEW_RELIC_SERVERLESS_MODE_ENABLED=CFG.NEW_RELIC_SERVERLESS_MODE_ENABLED,
         NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=CFG.NEW_RELIC_DISTRIBUTED_TRACING_ENABLED,
         PROFILING_ENABLED=CFG.PROFILING_ENABLED,
+        DEPLOY_DOMAIN=CFG.DEPLOY_DOMAIN,
     )
     return sep.join([
         f'{key}={value}' for key, value in dict(envs, **kwargs).items()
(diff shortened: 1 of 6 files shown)


dodo.py
```diff
@@ -319,7 +319,7 @@ def task_dynalite():
     dynalite db for testing and local runs: start, stop
     '''
     cmd = f'{DYNALITE} --port {CFG.DYNALITE_PORT}'
-    msg = f'dyanlite server started on {CFG.DYNALITE_PORT} logging to {CFG.DYNALITE_FILE}'
+    msg = f'dynalite server started on {CFG.DYNALITE_PORT} logging to {CFG.DYNALITE_FILE}'
```
bowlofstew: Minor type correction while in the neighborhood.  Feel free to call it out and I will revert and follow-up with a separate PR.

dodo.py
```diff
@@ -399,6 +399,48 @@ def task_yarn():
...
+            LongRunning(f'nohup {cmd} > /dev/null &'),
+            f'locust -f subhub/tests/performance/locustfile.py --host=http://localhost:{FLASK_PORT}'
+        ]
+    }
+
+def task_remote_perf():
+    '''
+    run locustio performance tests
+    '''
+    REMOTE_HOST='https://qa.fxa.mozilla-subhub.app'
```
bowlofstew: I happily extract this into a configuration as other variables should we suspect to leverage this in other environments in the future.

dodo.py
scottidler: that's fine; it was hard to see the diff

PR #137, opened 2019-07-17
Initial Behave Testing

bowlofstew:
After hearing in retro about the team wanting to investigate the utilization of the `behave` framework, I threw out an initial pull request with a proof of concept.  I did this as I have been using it for years and also that I could not sleep last night due to copious amounts of caffeine being ingested after 7 PM.

If the team wants, we can do a demo of it whenever you want.  I have also looped in Jorge from the ticket, [COPS-109](https://jira.mozilla.com/browse/COPS-109), in case he wants to run with it.

### DIFF
--- a/.idea/runConfigurations/behave.xml
+++ b/.idea/runConfigurations/behave.xml
@@ -1,0 +1,16 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="behave" type="PyBehaveRunConfigurationType" factoryName="Behave">
+    <module name="moz-subhub" />
+    <option name="INTERPRETER_OPTIONS" value="" />
+    <option name="PARENT_ENVS" value="true" />
+    <option name="SDK_HOME" value="$PROJECT_DIR$/venv/bin/python" />
+    <option name="WORKING_DIRECTORY" value="" />
+    <option name="IS_MODULE_SDK" value="false" />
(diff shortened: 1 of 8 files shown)


.idea/runConfigurations/behave.xml
```diff
@@ -0,0 +1,16 @@
+<component name="ProjectRunConfigurationManager">
```
sidler-mozilla: ugh,let's not check in your personal .idea/blah files

sidler-mozilla: xml, puke

bowlofstew: I know, I know.  I did this as it actually has a use and there was a `.gitignore` run to explicitly ignore files in this directory which required me to `git add -f .idea/runConfigurations/behave.xml`.  I did this as the value is that you can run these feature files under behave after [attaching to the process]()https://www.jetbrains.com/help/pycharm/attaching-to-local-process.html) being ran from the command, `doit local`.  This allows for easier debugging than good ol' `gdb python subhub/app.py` in the case of a failing `behave` test.  This being said is an interesting talking point as to how we want to bias developers towards tooling.  I happily remove it but just wanted to articulate my rationale as to why.

sidler-mozilla: ok let's discuss next week

bowlofstew: Cool!

PR #144, opened 2019-07-18
improve test coverage for secrets.py file

sidler-mozilla:
- mocked boto3.client used to retrieve aws secrets
- removed BinarySecret path
- added exception if SecretString not found
- added test for happy path to getting SecretString
- added test for SecretStringMissingError
- coverage 47% -> 87%
- the remaining 2 lines are hard to hit
- hitting the conditional during import is possible
  but caused issues with other tests (not worth it)


bowlofstew:
This looks awesome, nice addition sir!

PR #145, opened 2019-07-19
improved app.py test coverage; decorators remain

sidler-mozilla:


marty331:
I would say to push this to the QA environment and let's make sure all of the routes are working as expected prior to merging into master.

PR #147, opened 2019-07-19
Validate origin system

bowlofstew:
This pull request (PR) provides support to validate the origin system parameter in calls made to the API.  It will restrict such calls such that only configured calls are permitted origin_systems are permitted to be made.  The permitted origin_systems are stored in an AWS secret therefore this pull request requires an update thereof prior to deployment (dependency).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # SubHub
-payment subscription REST api for customers:
+
+[![Build Status](https://travis-ci.org/mozilla/subhub.svg?branch=master)](https://travis-ci.org/mozilla/subhub)
+
+Payment subscription REST api for customers:
 - FxA (Firefox Accounts)
 
 ## Required Software
--- a/subhub/cfg.py
+++ b/subhub/cfg.py
@@ -377,6 +377,13 @@ class AutoConfigPlus(AutoConfig):  # pylint: disable=too-many-public-methods
         return self("NEW_RELIC_DISTRIBUTED_TRACING_ENABLED", True)
 
     @property
+    def ALLOWED_ORIGIN_SYSTEMS(self):
+        """
+        ALLOWED_ORIGIN_SYSTEMS
+        """
+        return self("ALLOWED_ORIGIN_SYSTEMS", "").split(",")
+
+    @property
     def PROFILING_ENABLED(self):
         """
         PROFILING_ENABLED
--- a/subhub/customer.py
+++ b/subhub/customer.py
@@ -3,6 +3,7 @@ from stripe import Customer, Subscription
 import stripe
 from stripe.error import InvalidRequestError
 
+from subhub.cfg import CFG
 from subhub.exceptions import IntermittentError, ServerError
 from subhub.subhub_dynamodb import SubHubAccount
 from subhub.log import get_logger
@@ -18,6 +19,7 @@ def create_customer(
     origin_system: str,
     display_name: str,
 ) -> Customer:
+    _validate_origin_system(origin_system)
     # First search Stripe to ensure we don't have an unlinked Stripe record
     # already in Stripe
     customer = None
@@ -73,6 +75,7 @@ def existing_or_new_customer(
     origin_system: str,
     display_name: str,
 ) -> Customer:
+    _validate_origin_system(origin_system)
     customer = fetch_customer(subhub_account, user_id)
     if not customer:
         return create_customer(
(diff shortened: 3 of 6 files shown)


subhub/cfg.py
```diff
@@ -377,6 +377,13 @@ def NEW_RELIC_DISTRIBUTED_TRACING_ENABLED(self):
         """
         return self("NEW_RELIC_DISTRIBUTED_TRACING_ENABLED", True)
 
+    @property
+    def ALLOWED_ORIGIN_SYSTEMS(self):
+        """
+        ALLOWED_ORIGIN_SYSTEMS
+        """
+        return [x.lower() for x in (self("ALLOWED_ORIGIN_SYSTEMS", "").split(","))]
```
sidler-mozilla: why the `.lower()`?

subhub/customer.py
```diff
@@ -20,6 +21,10 @@ def create_customer(
 ) -> Customer:
     # First search Stripe to ensure we don't have an unlinked Stripe record
     # already in Stripe
+    if not origin_system.lower() in CFG.ALLOWED_ORIGIN_SYSTEMS:
+        raise InvalidRequestError(
+            message="Invalid origin_system provided", param=str(origin_system)
```
sidler-mozilla: the ALLOWED_ORIGIN_SYSTEMS are _not_ being passed to the Exception as we discussed.

subhub/customer.py
```diff
@@ -73,6 +78,10 @@ def existing_or_new_customer(
     origin_system: str,
     display_name: str,
 ) -> Customer:
+    if not origin_system.lower() in CFG.ALLOWED_ORIGIN_SYSTEMS:
```
sidler-mozilla: with the `.lower()` now we have these _uglier_ if conditionals.  What requirement is the lower() meeting?

subhub/tests/unit/test_payment_calls.py
```diff
@@ -7,13 +7,51 @@
...
 from subhub.log import get_logger
 
 logger = get_logger()
 
 
+def test_create_customer_invalid_origin_system():
+    """
+    GIVEN create a stripe customer
+    WHEN An invalid origin system is provided
+    THEN An exception should be raiseds
```
sidler-mozilla: typo

README.md
```diff
@@ -1,5 +1,8 @@
 # SubHub
-payment subscription REST api for customers:
+
+[![Build Status](https://travis-ci.org/mozilla/subhub.svg?branch=master)](https://travis-ci.org/mozilla/subhub)
```
sidler-mozilla: nice addition!

subhub/tests/unit/test_payment_calls.py
bowlofstew: Nice catch and EPIC FAIL!

subhub/customer.py
bowlofstew: I was taking case sensitivity out of the mix for callers and also application configuration for deployment agents (us).

subhub/cfg.py
bowlofstew: I was taking case sensitivity out of the mix for callers and also application configuration for deployment agents (us).

bowlofstew:
The last change set managed to break the build.  I am investigating now.

subhub/customer.py
```diff
@@ -134,3 +137,17 @@ def has_existing_plan(customer: Customer, plan_id: str) -> bool:
...
+
+
+def __validate_origin_system(origin_system: str):
+    """
+    This function validates a request's origin_system to validate that is permitted.  This
+    allows us to ensure that callers are permitted by configuration to interact with this application.
+    :param origin_system: The originating system in Mozilla
+    """
+    if origin_system not in CFG.ALLOWED_ORIGIN_SYSTEMS:
+        csv_origin_systems = ",".join(map(str, CFG.ALLOWED_ORIGIN_SYSTEMS))
```
sidler-mozilla: This seems unnecessary.  Doesn't the ALLOWED_ORIGIN_SYSTEMS property already return a list?  The __str__ or __repr__ version of the a list of strings isn't horrible to look at.
```
In [3]: import os

In [4]: os.environ["ALLOWED_ORIGIN_SYSTEMS"] = "Firefox,Salesforce"

In [5]: class Foo():
   ...:     @property
   ...:     def ALLOWED_ORIGIN_SYSTEMS(self):
   ...:         return os.environ["ALLOWED_ORIGIN_SYSTEMS"].split(',')
   ...: 

In [6]: foo = Foo()

In [7]: foo.ALLOWED_ORIGIN_SYSTEMS
Out[7]: ['Firefox', 'Salesforce']
```

sidler-mozilla: if not:
```
In [9]: ','.join(foo.ALLOWED_ORIGIN_SYSTEMS)
Out[9]: 'Firefox,Salesforce'
```

bowlofstew: Yup, I’ll update it.

bowlofstew: Yup, I just didn’t inline all of my changes after the lower() removal.  On it and sorry about that!

PR #155, opened 2019-07-20
added mpl2 header to the top of every .py file; auto to check|apply

sidler-mozilla:


marty331:
Everything works, I approve also.

PR #179, opened 2019-07-30
Process Missing Events

marty331:
- create serverless function for missing-events
- search for events that have not been processed
- send missing events to SF or FxA


services/missing-events/handler.py
```diff
@@ -0,0 +1,24 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+sys.path.append(dir_path)
+
+logger = get_logger()
+
+
+def handle(event, context):
+    try:
+        logger.info("handling event", subhub_event=event, context=context)
+        events_check.process_events(6)
```
bowlofstew: Why 6?

subhub/cfg.py
```diff
@@ -391,6 +391,13 @@ def ALLOWED_ORIGIN_SYSTEMS(self):
         """
         return self("ALLOWED_ORIGIN_SYSTEMS", "fake_origin1, fake_origin2").split(",")
 
+    @property
+    def PAYMENT_EVENT_LIST(self):
+        """"
+        PAYMENT_EVENT_LIST
+        """
+        return self("PAYMENT_EVENT_LIST", "test.system, test.event").split(",")
```
bowlofstew: Are these reasonable default values?

subhub/hub/verifications/events_check.py
```diff
@@ -0,0 +1,87 @@
...
+from flask import current_app
+import stripe
+
+from subhub.cfg import CFG
+from subhub.log import get_logger
+
+logger = get_logger()
+
+try:
+    app = create_app()
```
bowlofstew: Did you want to instrument this with XRay while at it?  You can just crib the other handler's middleware invocation.

subhub/cfg.py
marty331: This shows the pattern from Stripe, I thought that was sufficient.

subhub/hub/verifications/events_check.py
marty331: Yes, now that I have everything working I can add XRay.

PR #187, opened 2019-08-02
Initial caching workflow

bowlofstew:
This pull request (PR) adds support for caching of the most expensive call in the application.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/187)

### DIFF
--- a/subhub/sub/payments.py
+++ b/subhub/sub/payments.py
@@ -7,6 +7,7 @@
 
 from datetime import datetime
 
+import cachetools
 from stripe import Charge, Customer, Invoice, Plan, Product, Subscription
 from flask import g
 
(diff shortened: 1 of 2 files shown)


subhub/sub/payments.py
```diff
@@ -78,6 +79,11 @@ def list_all_plans() -> FlaskListResponse:
     List all available plans for a user to purchase.
     :return:
     """
+    return _get_all_plans(), 200
+
+
+@cachetools.cached(cachetools.TTLCache(10, 600))
+def _get_all_plans():
```
bowlofstew: This private method is tested implicitly via the direct call in `list_all_plans`.  If you want the additional unit tests to validate the delegation, just call it out.

PR #188, opened 2019-08-02
Prefer non-continuous deployment

bowlofstew:
This pull request (PR) represents a request by the team to leverage manual deployment from the master branch to the development environment in preference to continuous deployments from TravisCI.  This takes away knowledge of the encrypted AWS credentials from TravisCI as well which was a concern on this workflow.  This being said, a valid Stripe API key that can be used for our live or test environment which hosts customer data is still encrypted in TravisCI as it was prior to the CD change.  This being said, may present a far larger security risk than that of the AWS encrypted credentials.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/188)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,37 @@
 language: python
 dist: xenial
 python:
-  - "3.7"
+- 3.7
 env:
   global:
-    - AWS_DEFAULT_REGION: us-west-2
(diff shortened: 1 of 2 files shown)


bowlofstew:
As a note, leaving this open until we discuss as a team what we want to do around CD, etc.

PR #194, opened 2019-08-06
add the domain subtasks: create and delete

sidler-mozilla:


dodo.py
```diff
@@ -558,6 +558,48 @@ def task_pytest():
...
+        'actions': [
+            'echo "missing valid AWS credentials"',
+            'false',
+        ],
+        'uptodate': [
+            uptodate
+        ],
+    }
+
+def task_domain():
```
marty331: task_domain should not delete domain unless specified.

sidler-mozilla: impossible to achieve as it is currently configured

bowlofstew:
Do we need to add a task around the existing serverless task?  It seems heavyweight.

sidler-mozilla:
will rewrite another way

PR #195, opened 2019-08-07
COPS-327 - check_stripe_subscriptions

marty331:
- rename check_stripe_subscriptions to retrieve_stripe_subscriptions
- raise AttributeError if subscriptions node does not exist
- compare return value to submitted subscription id
- validate cancel_at_period_end is True
- update tests as appropriate

### DIFF
--- a/subhub/sub/payments.py
+++ b/subhub/sub/payments.py
@@ -103,15 +103,14 @@ def _get_all_plans():
     return stripe_plans
 
 
-def check_stripe_subscriptions(customer: Customer) -> list:
+def retrieve_stripe_subscriptions(customer: Customer) -> list:
     try:
-        logger.debug("check stripe subscriptions", subscriptions=customer.subscriptions)
         customer_subscriptions_data = customer.subscriptions
         customer_subscriptions = customer_subscriptions_data.get("data")
         return customer_subscriptions
-    except NameError as ne:
-        logger.error("error getting subscriptions", customer=customer, error=ne)
-        return []
+    except AttributeError as e:
+        logger.error("error getting subscriptions", customer=customer, error=e)
+        raise e
 
 
 def cancel_subscription(uid, sub_id) -> FlaskResponse:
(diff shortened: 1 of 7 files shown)


subhub/sub/payments.py
```diff
@@ -103,15 +103,14 @@ def _get_all_plans():
...
+def retrieve_stripe_subscriptions(customer: Customer) -> list:
     try:
-        logger.debug("check stripe subscriptions", subscriptions=customer.subscriptions)
         customer_subscriptions_data = customer.subscriptions
         customer_subscriptions = customer_subscriptions_data.get("data")
         return customer_subscriptions
-    except NameError as ne:
-        logger.error("error getting subscriptions", customer=customer, error=ne)
-        return []
+    except AttributeError as e:
```
sidler-mozilla: `AttributeError as ae`

subhub/sub/payments.py
```diff
@@ -103,15 +103,14 @@ def _get_all_plans():
...
+        logger.error("error getting subscriptions", customer=customer, error=e)
+        raise AttributeError()
```
sidler-mozilla: why not raise the actual error instead of instantiating a new AttributeError?

subhub/tests/unit/test_payments_mock.py
```diff
@@ -8,14 +8,14 @@
 import stripe
 
 from mockito import when, mock, unstub, ANY
-
+import pytest
```
sidler-mozilla: group `import blah` statements with others like `import stripe` above and group `from blah import blah` statements together in separate blocks; where you put `import pytest` doesn't follow this

marty331: Good catch, updated.

subhub/sub/payments.py
marty331: Good catch, updated.

subhub/sub/payments.py
marty331: Good catch, updated.

PR #211, opened 2019-08-13
Share common components

bowlofstew:
This pull request (PR) provides a single serverless release that releases shared components between 2 functions.  They are both bound to the FxA model presently as we need to discuss the complexities of adding support for a new tenant in the future and what that looks like.

An alternative approach to this method would be to use the prior 2 YAML files with a x-cloudstack reference but this approach albeit bound to a single tenant, is easier to understand.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/211)


bowlofstew:
I am keeping the branch open for now until we discuss then will delete.

PR #216, opened 2019-08-14
added DEPLOY_TO=prod guard; removed missing-events service with refac…

sidler-mozilla:
…tor of task_deploy


bowlofstew:
Were the files deleted outside of the change to `dodo.py` intentional with this change?

PR #222, opened 2019-08-22
COPS-620 Send Customer Data to Basket

marty331:
- Updated Customer Subscription Updated
  - Retrieve Invoice and Charge data to send complete records to basket
- Updated Customer Subscription Created
  - Retrieve Invoice info to send complete records to basket
- Removed unneeded hub Classes
- Updated tests to account for newly fetched data


subhub/hub/stripe/customer.py
```diff
@@ -198,29 +203,56 @@ def run(self):
...
                 )
                 logger.info("customer subscription cancel at period end", data=data)
-                routes = [StaticRoutes.FIREFOX_ROUTE, StaticRoutes.SALESFORCE_ROUTE]
+                routes = [StaticRoutes.SALESFORCE_ROUTE]
                 self.send_to_routes(routes, json.dumps(data))
             elif (
                 not self.payload.data.object.cancel_at_period_end
                 and self.payload.data.object.status == "active"
-                and not previous_attributes.get("cancel_at_period_end")
+                # and not previous_attributes.get("cancel_at_period_end")
```
bowlofstew: Comment still needed?

subhub/hub/stripe/customer.py
```diff
@@ -231,15 +263,22 @@ def run(self):
...
+                    current_period_start=self.payload.data.object.current_period_start,
+                    current_period_end=self.payload.data.object.current_period_end,
+                    invoice_number=invoice_number,
+                    brand=brand,
+                    last4=last4,
+                    charge=charge,
                 )
                 logger.info("customer subscription new recurring", data=data)
-                routes = [StaticRoutes.FIREFOX_ROUTE, StaticRoutes.SALESFORCE_ROUTE]
+                routes = [StaticRoutes.SALESFORCE_ROUTE]
```
bowlofstew: Was this part of the work here or just an in the neighborhood kind of thing?

subhub/hub/stripe/customer.py
```diff
@@ -254,3 +293,20 @@ def run(self):
...
+        return latest_invoice
+
+    def get_latest_charge(self, charge):
+        latest_charge = stripe.Charge.retrieve(id=charge)
+        logger.info("get latest charge", latest_charge=latest_charge)
+        if (
+            latest_charge.payment_method_details.card.last4 is None
+            or latest_charge.payment_method_details.card.brand is None
+        ):
+            self.get_latest_charge(charge)
```
bowlofstew: So the algorithm is that if for a payment type of credit card, if the last 4 digits of that payment type is not available or the payment credit card brand (VISA, Mastercard, etc) is not available, recursively call with the charge passed in?  I might be reading this incorrectly given the lack of sleep right now and all....

subhub/hub/stripe/customer.py
sidler-mozilla: yeah; let's delete

marty331: Actually I want it there as a referece, will most likely need to add a third scenario next week pending discussions between Stephen and Desiree.

subhub/hub/stripe/customer.py
marty331: FxA team asked for it to be removed.

subhub/hub/stripe/customer.py
marty331: Yes, the idea is if Stripe hasn't updated value prior to making this call (has happened already) then call it again till the value is updated so we can ship it to Basket.

subhub/hub/stripe/customer.py
```diff
@@ -98,13 +65,44 @@ def run(self):
 class StripeCustomerSubscriptionCreated(AbstractStripeHubEvent):
     def run(self):
         logger.info("customer subscription created", payload=self.payload)
+        invoice_number = None
+        invoice_id = None
+        charge = None
+        last4 = None
+        brand = None
```
sidler-mozilla: is there a point to setting these values to None?  If not, delete these assignments.

subhub/hub/stripe/customer.py
```diff
@@ -98,13 +65,44 @@ def run(self):
...
+                "latest invoice",
+                latest_invoice=latest_invoice.number,
+                charge=latest_invoice.charge,
+            )
+            invoice_number = latest_invoice.number
+            charge = latest_invoice.charge
+        except InvalidRequestError as e:
+            logger.error("Unable to retrieve invoice", error=e)
+            raise InvalidRequestError(
+                message="Unable to retrieve invoice", param=str(e)
```
sidler-mozilla: don't do this; instead use the `raise InvalidRequestError(blah) from e` pattern.  We spent a lot of time undoing stringifying exception instances.

subhub/hub/stripe/customer.py
```diff
@@ -98,13 +65,44 @@ def run(self):
...
+        try:
+            latest_charge = stripe.Charge.retrieve(id=charge)
+            logger.info(
+                "latest charge", last4=latest_charge.payment_method_details["card"]
+            )
+            last4 = latest_charge.payment_method_details["card"]["last4"]
+            brand = latest_charge.payment_method_details["card"]["brand"]
+        except InvalidRequestError as e:
+            logger.error("Unable to retrieve charge", error=e)
+            raise InvalidRequestError(message="Unable to retrieve charge", param=str(e))
```
sidler-mozilla: same as above `from e`

subhub/hub/stripe/customer.py
```diff
@@ -198,29 +203,56 @@ def run(self):
...
+                    charge = latest_invoice.charge
+                    logger.info("charge", charge=charge)
+                    latest_charge = self.get_latest_charge(charge)
+                    logger.info("latest charge", latest_charge=latest_charge)
+                    last4 = latest_charge.payment_method_details.card.last4
+                    brand = latest_charge.payment_method_details.card.brand
+                except InvalidRequestError as e:
+                    logger.error("Unable to gather data", error=e)
+                    raise InvalidRequestError(
+                        message="Unable to gather data", param=str(e)
```
sidler-mozilla: actually why are we not just `raise`ing the exception after logging it?  `raise e`

subhub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -124,46 +124,19 @@ def test_stripe_hub_customer_source_expiring(mocker):
...
+        "email": "user123@tester.com",
+        "customer_id": "cus_00000000000000",
+        "name": "Jon Tester",
+        "user_id": "user123",
     }
-    logger.info("created payload", data=data)
     basket_url = CFG.SALESFORCE_BASKET_URI + CFG.BASKET_API_KEY
     response = mock({"status_code": 200, "text": "Ok"}, spec=requests.Response)
-    when(boto3).client("sns", region_name=CFG.AWS_REGION).thenReturn(MockSnsClient)
+    when(boto3).client("sqs", region_name=CFG.AWS_REGION).thenReturn(MockSqsClient)
```
sidler-mozilla: Weird that this changed from Sns to Sqs; maybe explain that to us at some point.

marty331: good catch, changed.

subhub/hub/stripe/customer.py
bowlofstew: Understood.

subhub/hub/stripe/customer.py
bowlofstew: Gotcha so that solves the other issue in our tracker that I've seen floating around as part of this.

subhub/hub/stripe/customer.py
bowlofstew: So question on that algorithm then, how long are you willing to poll their API and should we have any wait period in between calls to avoid getting throttled from excessive calls?

marty331: Do you recommend doing a time.sleep(0.2) before calling again??  Something else?

PR #225, opened 2019-08-23
Resolve Webhook Issue

marty331:
- Stripe returning success as an error, now requires object with message
- Invoice, remove invoice_id and invoice_number on payment_failed


subhub/hub/stripe/controller.py
```diff
@@ -47,21 +47,21 @@ def view() -> tuple:
...
         sig_header = request.headers["Stripe-Signature"]
         event = stripe.Webhook.construct_event(payload, sig_header, CFG.HUB_API_KEY)
-        pipeline = StripeHubEventPipeline(event)
-        pipeline.run()
+        return {"message": ""}, 200
     except ValueError as e:
         # Invalid payload
         logger.error("ValueError", error=e)
-        return Response(status=400)
+        return {"message": e}, 400
```
bowlofstew: Do we need to update the Swagger for this?  https://github.com/mozilla/subhub/blob/22a4e08c1a5c59d9ddeab7b724e83d5df502c3ba/subhub/swagger.yaml#L498

subhub/hub/stripe/controller.py
```diff
@@ -47,21 +47,21 @@ def view() -> tuple:
...
     except Exception as e:
         logger.error("General Exception", error=e)
-        return Response(e, status=500)
-
-    return Response("Success", status=200)
+        return {"message": e}, 500
+    finally:
+        pipeline = StripeHubEventPipeline(event)
+        pipeline.run()
```
sidler-mozilla: did we remove the response "Success", 200 on purpose?  And if so what gets executed if it gets to that finally clause?

Issue #197, opened 2019-08-08
Use proper List typing


bbangert:
Multiple functions throughout subhub use `list` as the type for a parameter or result. They should however be using the `typing` modules `List` with the type of list it is:
https://docs.python.org/3/library/typing.html#type-aliases

For example, one function returns a list of subscriptions, which should be written as:
```python
from typing import List

def somefunc(...args) -> List[stripe.Subscription]:
    ....
```

mypy and PyCharm will then work correctly with the type information.

marty331:
Resolved with #228

PR #228, opened 2019-08-30
COPS-639 Implement Typing

marty331:
- Fix Issue #197
- Implement type checking for full application

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,3 +10,14 @@ testpaths = subhub
 
 [pytest]
 addopts = -p no:warnings
+
+[mypy]
+disallow_untyped_calls = True
+follow_imports = normal
+ignore_missing_imports = True
+python_version = 3.7
+strict_optional = True
+warn_no_return = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+warn_unreachable = True
--- a/subhub/cfg.py
+++ b/subhub/cfg.py
@@ -446,7 +446,7 @@ class AutoConfigPlus(AutoConfig):  # pylint: disable=too-many-public-methods
         """
         return self("DEPLOYED_BY", f"{self.USER}@{self.HOSTNAME}")
 
-    @property
+    @property  # type: ignore
     @lru_cache()
     def DEPLOYED_WHEN(self):
         """
--- a/subhub/customer.py
+++ b/subhub/customer.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 """Customer functions"""
+from typing import Union, Optional
 from stripe import Customer, Subscription
 import stripe
 from stripe.error import InvalidRequestError
@@ -32,7 +33,8 @@ def create_customer(
         if possible_customer.email == email:
             # If the userid doesn't match, the system is damaged.
             if possible_customer.metadata.get("userid") != user_id:
-                raise ServerError("customer email exists but userid mismatch")
+                customer_message = "customer email exists but userid mismatch"
+                raise ServerError(customer_message)
 
             customer = possible_customer
             # If we have a mis-match on the source_token, overwrite with the
@@ -62,10 +64,12 @@ def create_customer(
         uid=user_id, origin_system=origin_system, cust_id=customer.id
     )
 
-    if not subhub_account.save_user(db_account):
-        # Clean-up the Stripe customer record since we can't link it
-        Customer.delete(customer.id)
-        e = IntermittentError("error saving db record")
+    try:
+        new_user = subhub_account.save_user(db_account)
+        if not new_user:
+            # Clean-up the Stripe customer record since we can't link it
+            Customer.delete(customer.id)
+    except IntermittentError("error saving db record") as e:  # type: ignore
         logger.error("unable to save user or link it", error=e)
         raise e
     return customer
--- a/subhub/db.py
+++ b/subhub/db.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from typing import Optional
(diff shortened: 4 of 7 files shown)


subhub/customer.py
```diff
@@ -32,7 +32,9 @@ def create_customer(
         if possible_customer.email == email:
             # If the userid doesn't match, the system is damaged.
             if possible_customer.metadata.get("userid") != user_id:
-                raise ServerError("customer email exists but userid mismatch")
+                raise ServerError(  # type: ignore
```
sidler-mozilla: this is ugly; if black requires this, then maybe declare the str for the `customer email exists but userid mismatch` on a line a above

subhub/db.py
```diff
@@ -61,7 +61,7 @@ class Meta:
     def new_user(
         self, uid: str, origin_system: str, cust_id: Optional[str] = None
     ) -> SubHubAccountModel:
-        return self.model(
+        return self.model(  # type: ignore
```
sidler-mozilla: why are we ignoring the model type anyway? isn't this the kind of thing that we to check typing for?

subhub/db.py
```diff
@@ -70,7 +70,7 @@ def new_user(
 
     def get_user(self, uid: str) -> Optional[SubHubAccountModel]:
         try:
-            subscription_user = self.model.get(uid, consistent_read=True)
+            subscription_user: Any = self.model.get(uid, consistent_read=True)
```
sidler-mozilla: what is the benefit of typing to `Any`?

subhub/log.py
```diff
@@ -130,12 +132,14 @@ def _setup_once():
...
-def get_logger(logger_name=None):
+def get_logger(logger_name=None) -> structlog._config.BoundLoggerLazyProxy:
     global IS_CONFIGURED
     if not IS_CONFIGURED:
         IS_CONFIGURED = True
         _setup_once()
     if logger_name is None:
-        logger_name = inspect.currentframe().f_back.f_globals["__name__"]
+        logger_name = inspect.currentframe().f_back.f_globals[  # type: ignore
+            "__name__"
```
sidler-mozilla: this is awful|ugly; can we refactor so black doesnt make it look like this train-wreck?

subhub/cfg.py
```diff
@@ -446,7 +446,7 @@ def DEPLOYED_BY(self):
         """
         return self("DEPLOYED_BY", f"{self.USER}@{self.HOSTNAME}")
 
-    @property
+    @property  # type: ignore
```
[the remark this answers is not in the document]
marty331: Only line 449 was being called out by mypy

subhub/db.py
marty331: mypy throws a mismatch, it claims to be looking for SubHubAccountModel but returns SubHubAccountModel@427  (not sure the exact number here, going from memory), after spending more than an hour trying to determine a solution I finally went with ignoring this.

subhub/db.py
bowlofstew: It reads to me as this should respond with a `SubHubAccountModel`.

subhub/db.py
```diff
@@ -117,7 +117,7 @@ def mark_deleted(self, uid: str) -> bool:
             return False
 
 
-def _create_hub_model(table_name_, region_, host_):
+def _create_hub_model(table_name_, region_, host_) -> Any:
```
bowlofstew: `_create_hub_model` returns an optional model possibly?  I can get that this would error out so optional makes sense.

subhub/db.py
sidler-mozilla: I think because you defined the model twice no?

sidler-mozilla:
```
 sidler@76  ~/repos/mozilla/subhub   master  rg SubHubAccountModel
subhub/db.py
18:    class SubHubAccountModel(Model):
29:    return SubHubAccountModel
34:class SubHubAccountModel(Model):
47:        class SubHubAccountModel(Model):
59:        self.model = SubHubAccountModel
63:    ) -> SubHubAccountModel:
71:    def get_user(self, uid: str) -> Optional[SubHubAccountModel]:
80:    def save_user(user: SubHubAccountModel) -> bool:
```
see how the class is defined twice?

subhub/db.py
marty331: @scottidler @bowlofstew Changing Any to SubHubAccountModel causes the following error: `error: Incompatible types in assignment (expression has type "subhub.db.SubHubAccountModel@48", variable has type "subhub.db.SubHubAccountModel")`

subhub/db.py
marty331: @scottidler I agree in theory but the SubHubAccountModel defined on ln 35 is done so solely for type-checking, without it SubHubAccountModel is not able to be referenced as a return value.

subhub/db.py
sidler-mozilla: Yeah, like I say in the comment above, I think it might have something to do with that class being defined twice, which has never sat well with me anyway.  I could be wrong though, but we should figure it out.

subhub/db.py
marty331: mypy seems to be unable to distinguish the HubEventModel in the return from what is actually returned and something like `@47` (random number each time) so it then fails, it passes with Any.

setup.cfg
```diff
@@ -10,3 +10,13 @@ testpaths = subhub
...
+
+[mypy]
+disallow_untyped_calls = True
+follow_imports = normal
+ignore_missing_imports = True
+python_version = 3.7
+strict_optional = True
+warn_no_return = True
+warn_redundant_casts = True
+warn_unused_ignores = True
```
bowlofstew: `warn_unreachable`?

marty331: Added `warn_unreachable = True`  `allow_redefinition` is defaulted to false, which looks to be what would be wanted.  "Allows variables to be redefined with an arbitrary type, as long as the redefinition is in the same block and nesting level as the original definition."

PR #239, opened 2019-09-12
Make auth comparison less complex

claudijd:
When reviewing this code, I was a little worried about this auth pattern eventually happening if we cleaned this up later...

>>> "a" in "apple"
True
>>> "a" in ("apple",)
False

Meaning that if we drop the (,) part of the conditional just to simply down the line, we could create a future condition where an attacker would only need to substring match an API key to get successful auth.

This PR just makes the conditional logic as direct as it can be to ensure a true exact match.

### DIFF
--- a/src/shared/authentication.py
+++ b/src/shared/authentication.py
@@ -9,22 +9,27 @@ from shared.log import get_logger
 logger = get_logger()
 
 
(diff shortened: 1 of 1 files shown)


claudijd:
rebased|squashed|single-commit, done.

Issue #92, opened 2019-06-06
cannot use mutable object (dict) in keyword arguments


sidler-mozilla:
https://github.com/mozilla/subhub/blob/master/subhub/tests/unit/stripe/utils.py#L17-L22

Setting arguments using mutable default value is a bug:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

marty331:
subhub.test.unit.strip.utils should be refactored ln 30 & 33

PR #242, opened 2019-09-13
Using a mutable default value as an argument

bowlofstew:
This pull request (PR) addresses the remaining code called out in https://github.com/mozilla/subhub/issues/92.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/242)

### DIFF
--- a/src/hub/tests/unit/stripe/charge/test_stripe_charge.py
+++ b/src/hub/tests/unit/stripe/charge/test_stripe_charge.py
@@ -2,13 +2,16 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import os
 import mockito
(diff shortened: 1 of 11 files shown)


src/hub/tests/unit/stripe/utils.py
```diff
@@ -28,10 +28,10 @@ def run_event_process(event) -> Response:
 
 
 class MockSqsClient:
-    def list_queues(QueueNamePrefix={}) -> Any:  # type: ignore
+    def list_queues(QueueNamePrefix=None) -> Any:  # type: ignore
```
sidler-mozilla: i think you need to check if QueueNamePrefix is None and if so set to a default value; you have to keep the = None part though; same for the rest

bowlofstew: It's actually not used at all presently.

PR #247, opened 2019-09-18
COPS-626 Exponential backoff

marty331:
- Utilize exponential backoff for all calls to Stripe
- Implement Tenacity to handle exponential backoff
- Implement use of indempotency key for Stripe Post calls
- Positive and negative testing of all

### DIFF
--- a/src/shared/universal.py
+++ b/src/shared/universal.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import uuid
+
 
 def format_plan_nickname(product_name: str, plan_interval: str) -> str:
     interval_dict = {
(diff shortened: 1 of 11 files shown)


src/shared/universal.py
```diff
@@ -17,3 +19,7 @@ def format_plan_nickname(product_name: str, plan_interval: str) -> str:
         formatted_interval = plan_interval
 
     return f"{product_name} ({formatted_interval})".title()
+
+
+def get_indempotency_key() -> str:
```
bowlofstew: We will need to re-run performance around some of these cases.  This calls has performance implications.

marty331: So are you going to re-run perfomance? Or are you going to teach me how?  I'm good either way.

bowlofstew: After this lands, we can walk thru how to do it.

src/shared/vendor.py
```diff
@@ -0,0 +1,382 @@
...
+@retry(
+    wait=wait_exponential(multiplier=1, min=1, max=8),
+    stop=stop_after_attempt(4),
+    reraise=True,
+)
+def build_stripe_subscription(
+    customer_id: str, plan_id: str, idempotency_key: str
+) -> Subscription:
+    """
+    Create a new Stripe subscription for a given customer
```
bowlofstew: The only thing to note in documentation is that it appears to be inconsistent between these methods and others below.

PR #255, opened 2019-09-20
COPS-733 Subscribe to plan error

marty331:
- Explicitely call customer= in Subscribe.create
- Allow for using live test key for Stripe
- Added the missing env vars to the docker-compose.yml


src/sub/app.py
```diff
@@ -34,11 +34,12 @@ def is_docker() -> bool:
...
-        stripe.upload_api_base
-    ) = f"https://{CFG.STRIPE_MOCK_HOST}:{CFG.STRIPE_MOCK_PORT}"
-    logger.info("Stripe API URL", url=stripe.api_base)
-    stripe.verify_ssl_certs = False
+    if CFG.STRIPE_LOCAL is not True:
+        stripe.verify_ssl_certs = False
+        stripe.api_base = (
+            stripe.upload_api_base
+        ) = f"https://{CFG.STRIPE_MOCK_HOST}:{CFG.STRIPE_MOCK_PORT}"
+    logger.info("Stripe API URL", url=stripe.api_base, local=CFG.STRIPE_LOCAL)
```
bowlofstew: So this work is for the case when you are running in docker but without `stripe-mock`?  The alternative implementation for this could also be to `f"https://{CFG.STRIPE_MOCK_HOST}:{CFG.STRIPE_MOCK_PORT}"` to point the variables:

- STRIPE_MOCK_HOST
- STRIPE_MOCK_PORT

at `api.stripe.com` and 443 respectively.  This though for the reader, would require us to rename the variables from `*_MOCK_*` to an alternative.  This implementation would reduce the complexity to the configuration of the application.

marty331: Yes, this is intended to allow you to run with the actual Stripe API in place to test against how Stripe will actually respond.  Your statement above is truthy but that would require the dev to set/unset two variables vs one.  If you feel strongly about it then I can go with it.

bowlofstew: I get that and have no strong argument for or against this implementation.  I was just presenting an alternative approach.

PR #273, opened 2019-09-30
NewRelic Lambda Integration Deployment Scripts

bowlofstew:
This pull request (PR) adds script support that I created while working thru the NewRelic deployment workflows.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/273)


bin/new-relic/enable-streaming.sh
```diff
@@ -0,0 +1,21 @@
...
+    fi
+    if [ -d "$NR_LAMBDA_DIRECTORY" ]; then
+        cd nr-lambda
+        sh newrelic-cloud stream-lambda-logs $1
+    else
+        echo "Please run install.sh first."
+    fi
+}
+
+enable_streaming $1
```
sidler-mozilla: how will do this one or more?  shouldn't this be "$@"?

bin/new-relic/install.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+command -v git >/dev/null 2>&1 || {
```
sidler-mozilla: `hash <program>` is a cleaner, better way to test if an executable is available.

bowlofstew: Both are qualitative statements which I’d like to quantify here with pointing to the source code of both: https://github.com/bminor/bash/blob/master/builtins/command.def vs https://github.com/bminor/bash/blob/master/builtins/hash.def

Is this the case of the arg being used in cleaner?   Better: is the hashing of a previously traversed PATH location vs full PATH traversal on command.def?
I’m just trying to make the internal case for either/or here

bin/new-relic/enable-streaming.sh
bowlofstew: Sure.

PR #297, opened 2019-10-04
Resolve COPS-772

marty331:
- Split log statements in hub abstract and catch errors
- Log Firefox payload
- Update gitignore to ignore Vagrant and NR

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       dockerfile: Dockerfile
       args:
         LOCAL_FLASK_PORT: 5000
-        DYNALITE_PORT: 4567
+        DYNALITE_PORT: 8000
     environment:
         AWS_ACCESS_KEY_ID: "fake-id"
         AWS_SECRET_ACCESS_KEY: "fake-key"
@@ -40,7 +40,7 @@ services:
       - "5000:5000"
     depends_on:
       - base
-      - dynalite
+      - dynamodb
 
   hub:
     container_name: hub
@@ -51,7 +51,7 @@ services:
       dockerfile: Dockerfile
       args:
         LOCAL_FLASK_PORT: 5001
-        DYNALITE_PORT: 4567
+        DYNALITE_PORT: 8000
     environment:
         AWS_ACCESS_KEY_ID: "fake-id"
         AWS_SECRET_ACCESS_KEY: "fake-key"
@@ -73,12 +73,13 @@ services:
       - "5001:5001"
     depends_on:
       - base
-      - dynalite
+      - dynamodb
 
-  dynalite:
-    build: https://github.com/vitarn/docker-dynalite.git
+  dynamodb:
+    image: amazon/dynamodb-local:latest
+    container_name: dynamodb-local
     ports:
-      - 4567
+      - 8000
 
   stripe:
     image: stripemock/stripe-mock:latest
--- a/dodo.py
+++ b/dodo.py
@@ -77,6 +77,11 @@ def envs(sep=' ', **kwargs):
         SUPPORT_API_KEY=CFG.SUPPORT_API_KEY,
         PROJECT_NAME=CFG.PROJECT_NAME,
         VERSION=CFG.VERSION,
+        DYNALITE_PORT=CFG.DYNALITE_PORT,
+        LOCAL_FLASK_PORT=CFG.LOCAL_FLASK_PORT,
+        LOCAL_HUB_FLASK_PORT=CFG.LOCAL_HUB_FLASK_PORT,
+        STRIPE_LOCAL=CFG.STRIPE_LOCAL,
+
     )
     return sep.join([
         f'{key}={value}' for key, value in sorted(dict(envs, **kwargs).items())
@@ -464,8 +469,8 @@ def task_local():
 
 def task_yarn():
     '''
-    Install packages from package.json into the node_modules directory.  Yarn will attempt 
-    on non-OSX operating systems to check/attempt to install fsevents. 
+    Install packages from package.json into the node_modules directory.  Yarn will attempt
+    on non-OSX operating systems to check/attempt to install fsevents.
     This is evidenced in the log files by the messages:
         info fsevents@2.0.7: The platform "linux" is incompatible with this module.
         info "fsevents@2.0.7" is an optional dependency and failed compatibility check. Excluding it from installation.
@@ -763,5 +768,16 @@ def task_tidy():
         'actions': [
             'rm -rf ' + ' '.join(TIDY_FILES),
             'find . | grep -E "(__pycache__|\.pyc$)" | xargs rm -rf',
+        ],
+    }
+
+def task_dynamo():
+    return {
+        'task_dep': [
+            'check',
+            'tar'
+        ],
+        'actions': [
+            f'env {envs()} docker-compose up dynamodb'
         ],
     }
--- a/src/app_requirements.txt
+++ b/src/app_requirements.txt
@@ -1,8 +1,8 @@
 # requirements for hub.application to run
 # do not add testing reqs or automation reqs here
 attrdict==2.0.1
-boto3==1.9.184
-botocore==1.12.184
+boto3==1.9.245
+botocore==1.12.245
 certifi==2019.6.16
 chardet==3.0.4
 colorama==0.4.1
@@ -22,7 +22,7 @@ newrelic==4.20.1.121
 openapi-spec-validator==0.2.7
 pathlib==1.0.1
 psutil==5.6.3
-pynamodb==3.4.1
+pynamodb==4.0.0
 python-dateutil==2.8.0
 python-decouple==3.1
 PyYAML==5.1.1
--- a/src/hub/app.py
+++ b/src/hub/app.py
@@ -5,7 +5,6 @@
 import os
 import sys
 
-import logging
 import connexion
 import stripe
 
@@ -15,7 +14,7 @@ from flask import request
 
 from shared import secrets
 from shared.exceptions import SubHubError
-from shared.db import HubEvent
+from shared.db import HubEvent, SubHubDeletedAccount
 from shared.headers import dump_safe_headers
 from shared.cfg import CFG
 from shared.log import get_logger
@@ -57,15 +56,23 @@ def server_stripe_card_error(e):
     return jsonify({"message": f"{e.user_message}", "code": f"{e.code}"}), 402
 
 
-def is_container() -> bool:
-    import requests
+def is_docker() -> bool:
+    path = "/proc/self/cgroup"
+    return (
+        os.path.exists("/.dockerenv")
+        or os.path.isfile(path)
+        and any("docker" in line for line in open(path))
+    )
 
-    try:
-        requests.get(f"http://dynalite:{CFG.DYNALITE_PORT}")
-        return True
-    except Exception as e:
-        pass
-    return False
+
+if is_docker():
+    stripe.log = "DEBUG"
+    if CFG.STRIPE_LOCAL is not True:
+        stripe.verify_ssl_certs = False
+        stripe.api_base = (
+            stripe.upload_api_base
+        ) = f"https://{CFG.STRIPE_MOCK_HOST}:{CFG.STRIPE_MOCK_PORT}"
+    logger.info("Stripe API URL", url=stripe.api_base, local=CFG.STRIPE_LOCAL)
 
 
 def create_app(config=None):
@@ -73,8 +80,8 @@ def create_app(config=None):
     logger.info("creating flask app", config=config)
     region = "localhost"
     host = f"http://localhost:{CFG.DYNALITE_PORT}"
-    if is_container():
-        host = f"http://dynalite:{CFG.DYNALITE_PORT}"
+    if is_docker():
+        host = f"http://dynamodb:{CFG.DYNALITE_PORT}"
     stripe.api_key = CFG.STRIPE_API_KEY
     if CFG.AWS_EXECUTION_ENV:
         region = "us-west-2"
@@ -85,9 +92,17 @@ def create_app(config=None):
     app.add_api("swagger.yaml", pass_context_arg_name="request", strict_validation=True)
 
     app.app.hub_table = HubEvent(table_name=CFG.EVENT_TABLE, region=region, host=host)
+    app.app.subhub_deleted_users = SubHubDeletedAccount(
+        table_name=CFG.DELETED_USER_TABLE, region=region, host=host
+    )
 
     if not app.app.hub_table.model.exists():
         app.app.hub_table.model.create_table(
+            read_capacity_units=1, write_capacity_units=1, wait=True
+        )
+
+    if not app.app.subhub_deleted_users.model.exists():
+        app.app.subhub_deleted_users.model.create_table(
             read_capacity_units=1, write_capacity_units=1, wait=True
         )
 
@@ -125,6 +140,7 @@ def create_app(config=None):
         logger.debug("Request headers", headers=headers)
         logger.debug("Request body", body=request.get_data())
         g.hub_table = current_app.hub_table
+        g.subhub_deleted_users = current_app.subhub_deleted_users
         g.app_system_id = None
         if CFG.PROFILING_ENABLED:
             if "profile" in request.args and not hasattr(sys, "_called_from_test"):
@@ -151,4 +167,4 @@ if __name__ == "__main__":
     app = create_app()
     app.debug = True
     app.use_reloader = True
-    app.run(host="0.0.0.0", port=CFG.LOCAL_FLASK_PORT)
+    app.run(host="0.0.0.0", port=CFG.LOCAL_HUB_FLASK_PORT)
--- a/src/hub/routes/abstract.py
+++ b/src/hub/routes/abstract.py
@@ -17,16 +17,27 @@ class AbstractRoute(ABC):
 
     def report_route(self, payload: dict, sent_system: str) -> None:
         logger.info("report route", payload=payload, sent_system=sent_system)
-        existing = flask.g.hub_table.get_event(payload["event_id"])
+        if payload.get("event_id"):
+            event_id = payload["event_id"]
+        else:
+            event_id = payload["eventId"]
+        existing = flask.g.hub_table.get_event(event_id)
         if not existing:
-            created = flask.g.hub_table.new_event(
-                event_id=payload["event_id"], sent_system=sent_system
+            created_event = flask.g.hub_table.new_event(
+                event_id=event_id, sent_system=sent_system
             )
-            saved = flask.g.hub_table.save_event(created)
-            logger.info("new event", created=created, saved=saved)
+            saved = flask.g.hub_table.save_event(created_event)
+            try:
+                logger.info("new event created", created_event=created_event)
+            except Exception as e:
+                logger.error("Error logging created", error=e)
+            try:
+                logger.info("new event saved", saved=saved)
+            except Exception as e:
+                logger.error("Error logging saved", error=e)
         else:
             updated = flask.g.hub_table.append_event(
-                event_id=payload["event_id"], sent_system=sent_system
+                event_id=event_id, sent_system=sent_system
             )
             logger.info("updated event", existing=existing, updated=updated)
 
--- a/src/hub/routes/pipeline.py
+++ b/src/hub/routes/pipeline.py
@@ -5,6 +5,7 @@
 from hub.routes.firefox import FirefoxRoute
 from hub.routes.salesforce import SalesforceRoute
 from hub.routes.static import StaticRoutes
+from hub.shared.exceptions import UnsupportedStaticRouteError, UnsupportedDataError
 
 
 class RoutesPipeline:
@@ -19,4 +20,18 @@ class RoutesPipeline:
             elif r == StaticRoutes.FIREFOX_ROUTE:
                 FirefoxRoute(self.data).route()
             else:
-                raise Exception("We do no support " + str(r))
+                raise UnsupportedStaticRouteError(r, StaticRoutes)
+
+
+class AllRoutes:
+    def __init__(self, messages_to_routes) -> None:
+        self.messages_to_routes = messages_to_routes
+
+    def run(self) -> None:
+        for m in self.messages_to_routes:
+            if m["type"] == "firefox_route":
+                FirefoxRoute(m["data"]).route()
+            elif m["type"] == "salesforce_route":
+                SalesforceRoute(m["data"]).route()
+            else:
+                raise UnsupportedDataError(m, m["type"], StaticRoutes)
--- a/src/hub/tests/unit/stripe/customer/test_stripe_customer.py
+++ b/src/hub/tests/unit/stripe/customer/test_stripe_customer.py
@@ -9,13 +9,17 @@ import boto3
 import flask
 import stripe
 import requests
+import connexion
 
+from flask import g
 from mock import patch
 from mockito import when, mock, unstub
 
+from hub.app import create_app
 from hub.shared.tests.unit.utils import run_test, MockSqsClient, MockSnsClient
 from hub.shared.cfg import CFG
 from shared.log import get_logger
+
 
 logger = get_logger()
 
@@ -28,6 +32,15 @@ def run_customer(mocker, data, filename):
     flask.g.return_value = ""
 
     run_test(filename, cwd=CWD)
+
+
+def test_subhub():
+    """
+    Create an instance of the hub app and validate it is a FlaskApp
+    """
+    app = create_app()
+    logger.debug("g", g=dir(g))
+    assert isinstance(app, connexion.FlaskApp)
 
 
 def test_stripe_hub_customer_created(mocker):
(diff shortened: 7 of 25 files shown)


src/hub/routes/pipeline.py
```diff
@@ -19,4 +19,18 @@ def run(self) -> None:
             elif r == StaticRoutes.FIREFOX_ROUTE:
                 FirefoxRoute(self.data).route()
             else:
-                raise Exception("We do no support " + str(r))
+                raise Exception(f"We do not support " + {r})
```
sidler-mozilla: We should create our own Exception based error class, and then raise that.

sidler-mozilla: ie src/shared/exceptions.py

src/hub/vendor/customer.py
```diff
@@ -143,9 +143,41 @@ def run(self) -> None:
                 last4=last4,
                 charge=charge,
             )
+            # NOTE:
+            # Firefox's route has a particular casing requirement
+            # Salesforce's route doesn't care.
```
sidler-mozilla: Makes we should make them the same so there is less confusion when trying to cross correlate them.

src/hub/vendor/customer.py
```diff
@@ -143,9 +143,41 @@ def run(self) -> None:
...
+                        if k in data
+                    )
+                    payload = {"type": route, "data": json.dumps(subset)}
+                    subsets.append(payload)
+                    logger.info("subset", subset=subset)
+
+                    logger.debug("sent to", key=route)
+                except Exception as e:
+                    # log something and maybe change the exception type.
+                    logger.error("projection exception", error=str(e))
```
sidler-mozilla: be consistent; all of the other `error=e` occurrences in the codebase don't require str(); are you sure this isn't the same?

src/hub/routes/firefox.py
```diff
@@ -23,11 +23,12 @@ def route(self) -> None:
             sns_client = boto3.client("sns", region_name=CFG.AWS_REGION)
             response = sns_client.publish(
                 TopicArn=CFG.TOPIC_ARN_KEY,
-                Message=json.dumps({"default": json.dumps(self.payload)}),
+                Message=json.dumps({"default": self.payload}),
```
sidler-mozilla: this means that we will be passing json encoded strings, right?  why?

src/hub/routes/abstract.py
```diff
@@ -17,16 +17,29 @@ def __init__(self, payload) -> None:
...
+                event_id=event_id, sent_system=sent_system
             )
-            saved = flask.g.hub_table.save_event(created)
-            logger.info("new event", created=created, saved=saved)
+            saved = flask.g.hub_table.save_event(created_event)
+            try:
+                logger.info("new event created", created_event=created_event)
+
+            except Exception as e:
+                logger.error("Error logging created", error=str(e))
```
sidler-mozilla: drop str()

src/hub/routes/abstract.py
```diff
@@ -17,16 +17,29 @@ def __init__(self, payload) -> None:
...
+            try:
+                logger.info("new event saved", saved=saved)
+            except Exception as e:
+                logger.error("Error logging saved", error=str(e))
```
sidler-mozilla: drop str()

src/hub/routes/abstract.py
```diff
@@ -17,16 +17,29 @@ def __init__(self, payload) -> None:
 
     def report_route(self, payload: dict, sent_system: str) -> None:
         logger.info("report route", payload=payload, sent_system=sent_system)
-        existing = flask.g.hub_table.get_event(payload["event_id"])
+        event_id = None
+        if payload.get("event_id"):
```
sidler-mozilla: why do we have to have this ugly if block for `event_id` and `eventId`?

src/hub/routes/abstract.py
```diff
@@ -17,16 +17,29 @@ def __init__(self, payload) -> None:
...
-                event_id=payload["event_id"], sent_system=sent_system
+            created_event = flask.g.hub_table.new_event(
+                event_id=event_id, sent_system=sent_system
             )
-            saved = flask.g.hub_table.save_event(created)
-            logger.info("new event", created=created, saved=saved)
+            saved = flask.g.hub_table.save_event(created_event)
+            try:
+                logger.info("new event created", created_event=created_event)
+
```
sidler-mozilla: remove pointless blank line

.pre-commit-config.yaml
```diff
@@ -7,7 +7,6 @@ repos:
     -   id: check-added-large-files
     -   id: check-json
     -   id: detect-private-key
-    -   id: double-quote-string-fixer
```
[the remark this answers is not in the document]
marty331: This did not come in on my PR.

src/hub/vendor/customer.py
marty331: I disagree, FxA team is not bending on this and Basket team has everything working, so I feel we should leave as it is and continue forward.

src/hub/vendor/customer.py
marty331: Good point, should be.

src/hub/routes/firefox.py
marty331: This is a requirement from FxA.

src/hub/routes/abstract.py
marty331: done.

src/hub/routes/abstract.py
marty331: done

src/hub/routes/abstract.py
marty331: done

src/hub/routes/abstract.py
marty331: depending on the event type either event_id or eventId may be sent, this is needed to support FxA and Basket.

src/hub/vendor/customer.py
marty331: done.

.pre-commit-config.yaml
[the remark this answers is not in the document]
sidler-mozilla: @jackiemunroe or @bowlofstew ?

src/hub/routes/firefox.py
sidler-mozilla: Let's add a comment inline with the code saying that this is a requirement from them then; will remind us in the future

.pre-commit-config.yaml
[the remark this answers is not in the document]
bowlofstew: Did we want to use the `--skip-string-normalization` option then to allow this?

Reference: https://github.com/psf/black/issues/118

src/hub/routes/firefox.py
marty331: Done.

src/hub/routes/pipeline.py
marty331: done, thanks @scottidler

src/hub/vendor/abstract.py
```diff
@@ -47,3 +53,20 @@ def run(self) -> None:
...
+                payload = {"type": route, "data": json.dumps(subset)}
+                subsets.append(payload)
+                logger.info("subset", subset=subset)
+
+                logger.debug("sent to", key=route)
+            except Exception as e:
+                # log something and maybe change the exception type.
+                logger.error("projection exception", error=e)
+            finally:
+                self.send_to_all_routes(subsets)
```
jackiemunroe: Reading through this, it looks like this call should be outside of the for loop since it is defined above the logic. It appears that through each iteration data is appended to subsets, so it seems like the idea is for all of the subset data should be sent at once rather than at the end of each iteration. Is that the case?

src/hub/routes/abstract.py
```diff
@@ -17,16 +17,27 @@ def __init__(self, payload) -> None:
...
         if not existing:
-            created = flask.g.hub_table.new_event(
-                event_id=payload["event_id"], sent_system=sent_system
+            created_event = flask.g.hub_table.new_event(
+                event_id=event_id, sent_system=sent_system
             )
-            saved = flask.g.hub_table.save_event(created)
-            logger.info("new event", created=created, saved=saved)
+            saved = flask.g.hub_table.save_event(created_event)
+            try:
```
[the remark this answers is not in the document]
marty331: This is still very much in trouble shooting mode.

src/app_requirements.txt
```diff
@@ -22,7 +22,7 @@ newrelic==4.20.1.121
 openapi-spec-validator==0.2.7
 pathlib==1.0.1
 psutil==5.6.3
-pynamodb==3.4.1
```
bowlofstew: This and the 2 above package version upgrades were required for the index and query support.

src/hub/app.py
```diff
@@ -85,12 +84,20 @@ def create_app(config=None):
...
+    )
 
     if not app.app.hub_table.model.exists():
         app.app.hub_table.model.create_table(
             read_capacity_units=1, write_capacity_units=1, wait=True
         )
 
+    if not app.app.subhub_deleted_users.model.exists():
+        app.app.subhub_deleted_users.model.create_table(
+            read_capacity_units=1, write_capacity_units=1, wait=True
```
bowlofstew: Not to do it now but we should revisit the read and write capacity as part of the database model review soon.

marty331: Sounds good.  I think we should try to do that early next week.

src/hub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -30,6 +33,15 @@ def run_customer(mocker, data, filename):
     run_test(filename, cwd=CWD)
 
 
+def test_subhub():
+    """
+    something
```
bowlofstew: Please just don't put in meaningless docstrings if they don't add value.

src/shared/db.py
```diff
@@ -176,12 +178,22 @@ def remove_from_db(self, uid: str) -> bool:
...
     origin_system = UnicodeAttribute()
     customer_status = UnicodeAttribute()
+    subscription_info = ListAttribute()
 
 
 def _create_deleted_account_model(table_name_, region_, host_) -> Any:
+    class CustomerIndex(GlobalSecondaryIndex):
+        class Meta:
+            index_name = "cust-index"
+            read_capacity_units = 2
```
bowlofstew: Again as a follow up, we should discuss this in a follow up around the database model.  It shouldn't though hold up this pull request.

src/shared/db.py
```diff
@@ -219,6 +238,19 @@ def get_user(self, uid: str) -> Optional[SubHubDeletedAccountModel]:
...
             return None
 
+    def find_by_uid(self, uid: str):
+        items = [item for item in self.model.query(uid)]
+        return items
+
+    def find_by_cust(self, customer_id: str) -> Optional[SubHubDeletedAccountModel]:
+        logger.info("find cust start", customer_id=customer_id)
+        for item in self.model.cust_index.query(customer_id):
+            # for item in self.model.query("user_id", self.model.cust_id == customer_id):
```
bowlofstew: Remove commented out code.

src/hub/app.py
sidler-mozilla: create Jira Story for it please.  Now before we forget.

src/hub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -9,9 +9,12 @@
 import flask
 import stripe
 import requests
+import connexion
+from flask import g
 
 from mock import patch
 from mockito import when, mock, unstub
+from hub.app import create_app
```
sidler-mozilla: group our `from something import something` code together, separate from external libraries like `mock` and `mockito`

src/shared/db.py
```diff
@@ -2,10 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from typing import Optional, Any, List, cast, Type, TYPE_CHECKING, Union
+from typing import Optional, Any, List, Dict, cast, Type, TYPE_CHECKING, Union
```
sidler-mozilla: these two blocks of external library imports should be joined into one single block

src/shared/db.py
sidler-mozilla: Story?

src/shared/db.py
```diff
@@ -219,6 +238,19 @@ def get_user(self, uid: str) -> Optional[SubHubDeletedAccountModel]:
...
+        items = [item for item in self.model.query(uid)]
+        return items
+
+    def find_by_cust(self, customer_id: str) -> Optional[SubHubDeletedAccountModel]:
+        logger.info("find cust start", customer_id=customer_id)
+        for item in self.model.cust_index.query(customer_id):
+            # for item in self.model.query("user_id", self.model.cust_id == customer_id):
+            logger.info("find cust", item=item, user_id=item.user_id)
+            return item
+        logger.info("find cust did not find any results")
```
sidler-mozilla: this message is not very helpful if it doesn't mention what it was searching for

src/hub/tests/unit/stripe/customer/test_stripe_customer.py
marty331: done

src/shared/db.py
marty331: only one line and cleaned up.

src/shared/db.py
marty331: Done

src/shared/db.py
marty331: https://jira.mozilla.com/browse/COPS-784

src/hub/vendor/customer.py
```diff
@@ -36,6 +39,50 @@ def run(self) -> None:
...
+class StripeCustomerDeleted(AbstractStripeHubEvent):
+    def run(self) -> None:
+        logger.info("customer deleted", payload=self.payload)
+        customer_id = self.payload.data.object.id
+        uid = self.payload.data.object.metadata.userid
+        deleted_user = g.subhub_deleted_users.get_user(uid)
+        result = {}
+        for d in deleted_user.subscription_info:
+            for k in d.keys():
+                result[k] = result.get(k, 0) + d[k]
```
sidler-mozilla: what does this gem of logic do?

src/shared/db.py
marty331: You're right, removed.

src/shared/db.py
```diff
@@ -219,6 +238,15 @@ def get_user(self, uid: str) -> Optional[SubHubDeletedAccountModel]:
             logger.error("get user", uid=uid)
             return None
 
+    def find_by_uid(self, uid: str):
+        items = [item for item in self.model.query(uid)]
```
sidler-mozilla: why the list comprehension?

src/hub/vendor/customer.py
```diff
@@ -36,6 +39,50 @@ def run(self) -> None:
...
+        plan_amount = result.get("plan_amount", 0)
+        nicknames = " ".join([x["nickname"] for x in deleted_user.subscription_info])
+        subs = ",".join(
+            [str(x.get("subscription_id")) for x in deleted_user.subscription_info]
```
sidler-mozilla: `x["nickname"]` vs `x.get("subscription_id")` any particular reason that these are handled differently?  And why does one coerce to `str`?

marty331: Good question, applied .get to both and removed str.

src/hub/vendor/abstract.py
marty331: So the point of this function is to gather all the subsets and send to all routes, if none, 1 or more than 1 are gathered then the send_to_all_routes is called at the end of the try/except.  If nothing is added in subsets then it's sent with an empty list and no work will be done by the pipeline.

src/hub/vendor/customer.py
marty331: This gets the total amount for all subscriptions and sends to the routes so they can tell the customer how much they were paying in the emails.

src/shared/db.py
marty331: @bowlofstew I believe you wrote this one.

src/hub/vendor/abstract.py
jackiemunroe:
```
def customer_event_to_all_routes(self, data_projection, data):
        subsets = []
        for route in data_projection:
            try:
                logger.debug("sending to", key=route)
                subset = dict((k, data[k]) for k in data_projection[route] if k in data)
                payload = {"type": route, "data": json.dumps(subset)}
                subsets.append(payload)
                logger.info("subset", subset=subset)

                logger.debug("sent to", key=route)
            except Exception as e:
                # log something and maybe change the exception type.
                logger.error("projection exception", error=e)
        self.send_to_all_routes(subsets)
```

src/shared/db.py
sidler-mozilla: it appears as thought it could just `return self.model.query(uid)` and be done with it, no? @bowlofstew

bowlofstew: This goes back to the cardinality discussion earlier.  If our model can guarantee it is unique then we can simplify but that `query` call technically may be a 1 to many call per the data model.

src/hub/vendor/abstract.py
marty331: I see what you mean now.  Will update.

marty331: @jackiemunroe I moved the send_to_all_routes to the else on the for loop, there it will be called after the for loop completes.

src/shared/db.py
sidler-mozilla: Are you saying `model.query()` can return scalar OR list?  Even so can you explain how a list comprehension deals with whatever you think is happening please?

marty331: model.query() does return a list.  In the current setup we are not enforcing user_id to be unique, therefore there could potentially be multiple records with the same user_id.  This is preventable through code, obviously.

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       dockerfile: Dockerfile
       args:
         LOCAL_FLASK_PORT: 5000
-        DYNALITE_PORT: 4567
+        DYNALITE_PORT: 8000
```
sidler-mozilla: makes me think it should be in cfg.py

docker-compose.yml
```diff
@@ -73,12 +73,13 @@ services:
...
+      - dynamodb
 
-  dynalite:
-    build: https://github.com/vitarn/docker-dynalite.git
+  dynamodb:
+    image: amazon/dynamodb-local:latest
+    container_name: dynamodb-local
     ports:
-      - 4567
+      - 8000
```
sidler-mozilla: same

dodo.py
```diff
@@ -765,3 +765,10 @@ def task_tidy():
             'find . | grep -E "(__pycache__|\.pyc$)" | xargs rm -rf',
         ],
     }
+
+def task_dynamo():
```
sidler-mozilla: nothing depends on this task?

src/hub/app.py
```diff
@@ -57,24 +56,32 @@ def server_stripe_card_error(e):
...
 
-    try:
-        requests.get(f"http://dynalite:{CFG.DYNALITE_PORT}")
-        return True
-    except Exception as e:
-        pass
-    return False
+
+if is_docker():
+    stripe.log = "DEBUG"
```
sidler-mozilla: why the hardcoded value here?

src/hub/tests/unit/stripe/customer/test_stripe_customer.py
```diff
@@ -9,14 +9,18 @@
 import flask
 import stripe
 import requests
+import connexion
+from flask import g
```
sidler-mozilla: separate `from imports` from `imports` in external and internal code

src/app_requirements.txt
bowlofstew: We were required to upgrade this to support the new queries being used.

src/sub/messages.py
```diff
@@ -0,0 +1,44 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import json
+
```
sidler-mozilla: group `imports` of external code

src/sub/messages.py
```diff
@@ -0,0 +1,44 @@
...
+import boto3
+
+from abc import ABC
+
+from sub.shared.cfg import CFG
+from shared.log import get_logger
+
+from botocore.exceptions import ClientError
```
sidler-mozilla: group `from imports` of external code _above_ our code imports

src/app_requirements.txt
marty331: Is there a question here?

src/hub/vendor/abstract.py
```diff
@@ -47,3 +53,19 @@ def run(self) -> None:
...
+                logger.debug("sending to", key=route)
+                subset = dict((k, data[k]) for k in data_projection[route] if k in data)
+                payload = {"type": route, "data": json.dumps(subset)}
+                subsets.append(payload)
+                logger.info("subset", subset=subset)
+                logger.debug("sent to", key=route)
+            except Exception as e:
+                # log something and maybe change the exception type.
+                logger.error("projection exception", error=e)
+        else:
```
jackiemunroe: This `else` doesn't match up to anything and should be removed.

src/hub/vendor/customer.py
```diff
@@ -36,6 +39,52 @@ def run(self) -> None:
...
+        customer_id = self.payload.data.object.id
+        uid = self.payload.data.object.metadata.userid
+        deleted_user = g.subhub_deleted_users.get_user(uid, customer_id)
+        result = {}
+        for d in deleted_user.subscription_info:
+            for k in d.keys():
+                result[k] = result.get(k, 0) + d[k]
+        plan_amount = result.get("plan_amount", 0)
+        nicknames = " ".join(
+            [x.get("nickname") for x in deleted_user.subscription_info]
```
jackiemunroe: Does this refer to the Plan.nickname property? If so, this should be updated to use the `format_plan_nickname` method from the `shared.utils` file.

src/hub/app.py
bowlofstew: It allows logging of the Stripe API calls in debug mode to be able to accurately debug what is occurring between the Stripe API client and Stripe API.  I believe it has been in the code for some time as well.

src/sub/payments.py
```diff
@@ -165,47 +168,108 @@ def delete_customer(uid: str) -> FlaskResponse:
...
+            data_type=type(subscribed_customer),
         )
-        user = g.subhub_account.get_user(uid)
-        if deleted_customer and user is None:
-            return dict(message="Customer deleted successfully"), 200
-
+        for subs in subscribed_customer["subscriptions"]["data"]:
+            sub = dict(
+                plan_amount=subs.plan.amount,
+                nickname=subs.plan.nickname,
```
jackiemunroe: This needs to use the `format_plan_nickname` method

src/app_requirements.txt
bowlofstew: Nope.  I was merely calling out this for other readers.

src/hub/app.py
marty331: @scottidler  This is only for running locally and therefore should always be set to debug.

src/hub/vendor/abstract.py
marty331: I disagree, I purposefully put this in the else block in the for loop.  it is in the correct place.

src/hub/app.py
sidler-mozilla: If so, this seems like a strange place to set the value?  Always is this predicate answering function?  Strange to me.

src/hub/vendor/abstract.py
jackiemunroe: I thought that the `else` was only used if there was a `break` in a `for` loop?

marty331: Else will only be used if there is NOT a break in the for loop!

marty331: https://book.pythontips.com/en/latest/for_-_else.html#else-clause

src/hub/vendor/customer.py
jackiemunroe: https://github.com/mozilla/subhub/issues/311

src/sub/payments.py
jackiemunroe: https://github.com/mozilla/subhub/issues/311

dodo.py
marty331: @scottidler added depends, please review.

src/hub/tests/unit/stripe/customer/test_stripe_customer.py
marty331: @scottidler updated imports, please review.

src/sub/messages.py
marty331: @scottidler updated, please review.

src/sub/messages.py
marty331: @scottidler updated, please review.

src/sub/payments.py
marty331: @jackiemunroe updated, please review.

jackiemunroe: The nickname formatter uses the Product.name. The Plan.nickname should not be one of the parameters.

docker-compose.yml
sidler-mozilla: future issue will resolve

PR #300, opened 2019-10-06
Defect: Remove Dynalite Dependency

bowlofstew:
This pull request (PR) removes an outstanding issue that the subhub code has had for some time.  The issue is that there is an issue in which under certain circumstances that there is contention for the port used by the instantiated instance of Dynalite for the purposes of execution of unit tests against.  Additionally, a further but known defect is that the build system, doit and the test framework both attempt to managed the life cycle of the dynalite.

This PR is a free-time contribution and I would like to call out the following:

1. A black formatter, pre-commit hook was added as a few times, I did commit code that broke the build on the formatter check.  This pre-commit step ensure the black formatter is ran prior to code commit so in that sense, it is a superset of doit's ability to do this.  We should discuss removing this extraneous step from doit in the future to lower it's overall complexity given this.

2.  There exists duplicated code in 2 test fixtures to spin up dynalite for the sub and hub.  We may be able to abstract this into shared code but caveat emptor as this will potentially reduce the ability to call `tox --parallel auto` should we want to do this.  This is due to the abilities of both sub and hub to spin up dependent test fixtures on differing ports.  There is some additional work required to abstract this into a shared pytest fixture that is parallelizable.

3.  Changes the docker-compose.yml to leverage prebuilt docker images from dockerhub versus building from source.  This speeds up local development environment, build times.

4.  Caching the dynalite docker image in TravisCI needs to be investigated.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/300)

### DIFF
--- a/src/hub/app.py
+++ b/src/hub/app.py
@@ -57,24 +57,20 @@ def server_stripe_card_error(e):
     return jsonify({"message": f"{e.user_message}", "code": f"{e.code}"}), 402
 
 
-def is_container() -> bool:
-    import requests
-
-    try:
-        requests.get(f"http://dynalite:{CFG.DYNALITE_PORT}")
-        return True
-    except Exception as e:
-        pass
-    return False
+def is_docker() -> bool:
+    path = "/proc/self/cgroup"
+    return (
+        os.path.exists("/.dockerenv")
+        or os.path.isfile(path)
+        and any("docker" in line for line in open(path))
+    )
 
 
 def create_app(config=None):
     # configure_logger()
     logger.info("creating flask app", config=config)
     region = "localhost"
-    host = f"http://localhost:{CFG.DYNALITE_PORT}"
-    if is_container():
-        host = f"http://dynalite:{CFG.DYNALITE_PORT}"
+    host = f"http://dynalite:{CFG.DYNALITE_PORT}" if is_docker() else CFG.DYNALITE_URL
     stripe.api_key = CFG.STRIPE_API_KEY
     if CFG.AWS_EXECUTION_ENV:
         region = "us-west-2"
--- a/src/hub/tests/conftest.py
+++ b/src/hub/tests/conftest.py
@@ -1,15 +1,18 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
+import json
 import os
 import sys
 import signal
 import subprocess
 import logging
 
+import backoff
+import docker
 import psutil
 import pytest
+import requests
 import stripe
 from flask import g
 
@@ -19,17 +22,18 @@ from shared.log import get_logger
 
 logger = get_logger()
 
-ddb_process = None
+IMAGE = "vitarn/dynalite:latest"
+CONTAINER_FOR_TESTING_LABEL = "pytest_docker_log"
+
+
+def _docker_client():
+    return docker.DockerClient("unix://var/run/docker.sock", version="auto")
 
 
 def pytest_configure():
     """Called before testing begins"""
-    global ddb_process
     for name in ("boto3", "botocore"):
         logging.getLogger(name).setLevel(logging.CRITICAL)
-    if os.getenv("AWS_LOCAL_DYNAMODB") is None:
-        os.environ["AWS_LOCAL_DYNAMODB"] = f"http://127.0.0.1:{CFG.DYNALITE_PORT}"
-
     # Latest boto3 now wants fake credentials around, so here we are.
     os.environ["AWS_ACCESS_KEY_ID"] = "fake"
     os.environ["AWS_SECRET_ACCESS_KEY"] = "fake"
@@ -37,32 +41,57 @@ def pytest_configure():
     os.environ["ALLOWED_ORIGIN_SYSTEMS"] = "Test_system,Test_System,Test_System1"
     sys._called_from_test = True
 
-    # Locate absolute path of dynalite
-    dynalite = f"{CFG.REPO_ROOT}/node_modules/.bin/dynalite"
 
-    cmd = f"{dynalite} --port {CFG.DYNALITE_PORT}"
-    ddb_process = subprocess.Popen(
-        cmd, shell=True, env=os.environ, stdout=subprocess.PIPE
+def get_free_tcp_port():
+    import socket
+
+    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    tcp.bind(("", 0))
+    addr, port = tcp.getsockname()
+    logger.debug("port", port=port)
+    tcp.close()
+    return port
+
+
+def pull_image(image):
+    docker_client = docker.from_env()
+    response = docker_client.api.pull(image)
+    lines = [line for line in response.splitlines() if line]
+    pull_result = json.loads(lines[-1])
+    if "error" in pull_result:
+        raise Exception("Could not pull {}: {}".format(image, pull_result["error"]))
+
+
+@pytest.yield_fixture(scope="module", autouse=True)
+def dynamodb():
+    pull_image(IMAGE)
+    docker_client = docker.from_env()
+    host_port = get_free_tcp_port()
+    port_bindings = {4567: host_port}
+    host_config = docker_client.api.create_host_config(port_bindings=port_bindings)
+    container = docker_client.api.create_container(
+        image=IMAGE,
+        labels=[CONTAINER_FOR_TESTING_LABEL],
+        host_config=host_config,
+        ports=[4567],
     )
-    while 1:
-        line = ddb_process.stdout.readline()
-        if line.startswith(b"Listening"):
-            break
+    docker_client.api.start(container=container["Id"])
+    container_info = docker_client.api.inspect_container(container.get("Id"))
+    host_ip = container_info["NetworkSettings"]["Ports"]["4567/tcp"][0]["HostIp"]
+    host_port = container_info["NetworkSettings"]["Ports"]["4567/tcp"][0]["HostPort"]
+    yield f"http://{host_ip}:{host_port}"
+    docker_client.api.remove_container(container=container["Id"], force=True)
 
 
-def pytest_unconfigure():
-    del sys._called_from_test
-    global ddb_process
-    """Called after all tests run and warnings displayed"""
-    proc = psutil.Process(pid=ddb_process.pid)
-    child_procs = proc.children(recursive=True)
-    for p in [proc] + child_procs:
-        os.kill(p.pid, signal.SIGTERM)
-    ddb_process.wait()
+@backoff.on_exception(backoff.fibo, Exception, max_tries=8)
+def _check_container(url):
+    return requests.get(url)
 
 
 @pytest.fixture(autouse=True, scope="module")
-def app():
+def app(dynamodb):
+    os.environ["DYNALITE_URL"] = dynamodb
+    _check_container(dynamodb)
     app = create_app()
     with app.app.app_context():
         g.hub_table = app.app.hub_table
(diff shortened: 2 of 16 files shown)


src/shared/cfg.py
```diff
@@ -291,20 +291,17 @@ def LOCAL_FLASK_PORT(self):
         """
         return self("LOCAL_FLASK_PORT", 5000, cast=int)
 
+    @property
+    def DYNALITE_URL(self):
+        return self("DYNALITE_URL", "http://127.0.0.1:8000")
```
bowlofstew: A default item is debatable here.  Furthermore, no docstring was added as I can see no code documentation generation tool being used and the field naming is self evident.

sidler-mozilla: oh it is, lookie here; maybe a convo about how this is being used.

src/sub/tests/conftest.py
```diff
@@ -56,46 +61,71 @@ def get_file(filename, path=THIS_PATH, **overrides):
...
+    import socket
+
+    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    tcp.bind(("", 0))
+    addr, port = tcp.getsockname()
+    tcp.close()
+    return port
+
+
+def _docker_client():
```
sidler-mozilla: what is the benefit of this function?

src/sub/tests/conftest.py
```diff
@@ -27,9 +31,10 @@
 
 logger = get_logger()
 
-ddb_process = None
 THIS_PATH = os.path.join(os.path.realpath(os.path.dirname(__file__)))
 UID = str(uuid.uuid4())
+IMAGE = "vitarn/dynalite:latest"
```
sidler-mozilla: make me think it should be parameterized; perhaps this as a default value in cfg.py

src/sub/tests/conftest.py
```diff
@@ -27,9 +31,10 @@
...
+CONTAINERS_FOR_TESTING_LABEL = "pytest_docker_log"
```
sidler-mozilla: seems like another hardcoded configuration value; on top of that the variable and value naming are unclear

sidler-mozilla: Why "containers" plural?

src/sub/tests/conftest.py
```diff
@@ -56,46 +61,71 @@ def get_file(filename, path=THIS_PATH, **overrides):
...
     """Called before testing begins"""
-    global ddb_process
     for name in ("boto3", "botocore"):
         logging.getLogger(name).setLevel(logging.CRITICAL)
-    if os.getenv("AWS_LOCAL_DYNAMODB") is None:
-        os.environ["AWS_LOCAL_DYNAMODB"] = f"http://127.0.0.1:{CFG.DYNALITE_PORT}"
 
-    # Latest boto3 now wants fake credentials around, so here we are.
     os.environ["AWS_ACCESS_KEY_ID"] = "fake"
     os.environ["AWS_SECRET_ACCESS_KEY"] = "fake"
```
sidler-mozilla: not part of your code change, but i dont know that we need AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY anymore.  Those values are set in dodo.py and available as CFG values throughout our code.

src/sub/tests/conftest.py
```diff
@@ -56,46 +61,71 @@ def get_file(filename, path=THIS_PATH, **overrides):
...
+        host_config=host_config,
+        ports=[4567],
     )
-    while 1:
-        line = ddb_process.stdout.readline()
-        if line.startswith(b"Listening"):
-            break
+    docker_client.api.start(container=container["Id"])
+    container_info = docker_client.api.inspect_container(container.get("Id"))
+
```
sidler-mozilla: why the space in an otherwise solid block of code?

src/hub/tests/conftest.py
```diff
@@ -37,32 +45,61 @@ def pytest_configure():
...
+
+    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    tcp.bind(("", 0))
+    addr, port = tcp.getsockname()
+    tcp.close()
+    return port
+
+
+def _docker_client():
+    return docker.from_env()
```
sidler-mozilla: what is the value of this function?

src/hub/tests/conftest.py
```diff
@@ -37,32 +45,61 @@ def pytest_configure():
     os.environ["ALLOWED_ORIGIN_SYSTEMS"] = "Test_system,Test_System,Test_System1"
     sys._called_from_test = True
 
-    # Locate absolute path of dynalite
-    dynalite = f"{CFG.REPO_ROOT}/node_modules/.bin/dynalite"
 
-    cmd = f"{dynalite} --port {CFG.DYNALITE_PORT}"
-    ddb_process = subprocess.Popen(
-        cmd, shell=True, env=os.environ, stdout=subprocess.PIPE
+def get_free_tcp_port():
```
sidler-mozilla: what happens if the process hangs?  I will i know why port it ran on?  log it?

src/hub/tests/conftest.py
bowlofstew: https://sourcegraph.com/github.com/docker/docker-py/-/blob/docker/client.py#L43

src/hub/tests/conftest.py
```diff
@@ -37,32 +45,61 @@ def pytest_configure():
...
+
+@backoff.on_exception(backoff.fibo, Exception, max_tries=8)
+def _check_container(url):
+    return requests.get(url)
 
 
 @pytest.fixture(autouse=True, scope="module")
-def app():
+def app(dynamodb):
+    os.environ["DYNALITE_URL"] = dynamodb
```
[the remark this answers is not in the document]
bowlofstew: `dynamo` is not a string but fully qualified host and port based URL to the docker container.  It isn't a string but in the form of:

```
    host_ip = container_info["NetworkSettings"]["Ports"]["4567/tcp"][0]["HostIp"]
    host_port = container_info["NetworkSettings"]["Ports"]["4567/tcp"][0]["HostPort"]
    yield f"http://{host_ip}:{host_port}"
```
Given this is a dynamically created value based upon dynamic port lookup, I don't know the methodology by which we would make this a static configuration.

src/sub/tests/conftest.py
bowlofstew: It is called in multiple spots so it was a way to enforce the DRY principle.  I can remove if it is bothersome.

src/sub/tests/conftest.py
bowlofstew: I would opt to keep it here, extract a pytest fixture soon that is parallelizable that can have this in one place.  This would allow us to keep the configuration's surface area to a minimum for future maintainers.  If all are good with this approach, I have a todo item to do exactly this in my free-time at some point.

src/sub/tests/conftest.py
bowlofstew: Agreed.  I am going to leave that alone though on this pull request (PR) as it is functionally unrelated to the work.

src/shared/cfg.py
sidler-mozilla: heh; then we should delete them all; i did that so things like pylint wont complain; self-evident agreed though

src/hub/tests/conftest.py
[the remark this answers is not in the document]
sidler-mozilla: confused; youre hard-coding it in the code above; why not a default configuration?

src/hub/tests/unit/stripe/event/test_stripe_events.py
```diff
@@ -25,21 +26,27 @@
 
 
 def test_hours_back():
+    from hub.verifications.events_check import EventCheck
```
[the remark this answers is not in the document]
sidler-mozilla: still left unanswered

src/hub/tests/conftest.py
sidler-mozilla: ? i dont know what the code reference was supposed to mean to me; why have a private (starting with an '_') function to return the result of a function off of what appears to be a global `docker` object?  Why not just just the body of this function, inline at the call sites for _docker_client() ?

src/shared/cfg.py
sidler-mozilla: it's also bad practice to depart from prevailing coding style (in this case docstrs on all properties) without addressing the issue raise with every one of them.  Leads to messy code and lack of consistency throughout

src/sub/tests/conftest.py
sidler-mozilla: Not let it be a todo item, but rather add it as an item to the backlog in Jira

src/hub/tests/conftest.py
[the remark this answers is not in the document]
bowlofstew: host_ip and host_port aren't hard coded at all they are dynamically mapped.  The container port is hard coded as it is what is exposed by the docker container pulled down from dockerhub.

src/hub/tests/unit/stripe/event/test_stripe_events.py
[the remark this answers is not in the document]
bowlofstew: It does due to a create_app.hub_table call in the file.

src/hub/tests/unit/stripe/event/test_stripe_events.py
```diff
@@ -25,21 +26,27 @@
...
     event_check_class = EventCheck(hours_back=1)
     assert isinstance(
         event_check_class.get_time_h_hours_ago(hours_back=event_check_class.hours_back),
         int,
     )
 
 
-def test_process_missing_event():
+def test_process_missing_event(dynamodb):
+    from hub.verifications.events_check import EventCheck, process_events
```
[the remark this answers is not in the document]
bowlofstew: It is needed.

src/hub/tests/unit/stripe/event/test_stripe_events.py
```diff
@@ -25,21 +26,27 @@
...
+
     missing_event = "event.json"
     with open(os.path.join(CWD, missing_event)) as f:
         event_check = EventCheck(6)
         event_check.process_missing_event(json.load(f))
 
 
-def test_retrieve_events():
+def test_retrieve_events(dynamodb):
+    from hub.verifications.events_check import EventCheck, process_events
```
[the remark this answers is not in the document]
bowlofstew: Needed.

src/sub/tests/conftest.py
bowlofstew: resolved.

src/hub/app.py
```diff
@@ -72,7 +72,7 @@ def create_app(config=None):
     # configure_logger()
     logger.info("creating flask app", config=config)
     region = "localhost"
-    host = f"http://localhost:{CFG.DYNALITE_PORT}"
+    host = CFG.DYNALITE_URL
```
sidler-mozilla: ternary could be cool here?

src/sub/tests/conftest.py
sidler-mozilla: yeah generally im pissed if i go look for a function def and it is this trivial

src/shared/cfg.py
bowlofstew: In another work stream, I propose to add a pylint pre-commit hook.  As an option to that hook, I propose adding an option to `--disable` docstring warnings.  It is easy enough to add in.

src/hub/tests/conftest.py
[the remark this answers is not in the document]
sidler-mozilla: `os.environ["DYNALITE_URL"] = dynamodb` in this statement `dynamodb` is hard-coded; can you speak to that?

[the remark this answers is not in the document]
sidler-mozilla: Or does it come in as some object from somewhere?

src/sub/tests/conftest.py
bowlofstew: https://jira.mozilla.com/browse/COPS-790

src/sub/tests/conftest.py
bowlofstew: That is much akin to my feelings about the 0 value docstings in the cfg.py file.

src/shared/cfg.py
bowlofstew: Updated accordingly.

PR #315, opened 2019-10-15
Resolve Deletion Error

marty331:
- Correct vars when looping through deletedUser subscription info
- Add productId to SNS payload for FxA

### DIFF
--- a/src/hub/vendor/customer.py
+++ b/src/hub/vendor/customer.py
@@ -45,27 +45,15 @@ class StripeCustomerDeleted(AbstractStripeHubEvent):
         customer_id = self.payload.data.object.id
         uid = self.payload.data.object.metadata.userid
         deleted_user = g.subhub_deleted_users.get_user(uid, customer_id)
-        result = {}
-        for d in deleted_user.subscription_info:
-            for k in d.keys():
-                result[k] = result.get(k, 0) + d[k]
-        plan_amount = result.get("plan_amount", 0)
-
-        formatted_nicknames = list()
-        products = {}  # type: Dict
+        plan_amount = 0
+        current_period_end = None
+        current_period_start = None
+        nicknames = list()
         for user_plan in deleted_user.subscription_info:
-            try:
-                product = products[user_plan["product"]]
-            except KeyError:
-                product = stripe.Product.retrieve(user_plan["product"])
-                products[user_plan["product"]] = product
-
-            plan_nickname = format_plan_nickname(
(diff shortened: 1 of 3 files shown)


src/hub/vendor/customer.py
```diff
@@ -46,56 +46,38 @@ def run(self) -> None:
...
         result = {}
-        for d in deleted_user.subscription_info:
-            for k in d.keys():
-                result[k] = result.get(k, 0) + d[k]
-        plan_amount = result.get("plan_amount", 0)
-
-        formatted_nicknames = list()
-        products = {}  # type: Dict
+        # for d in deleted_user.subscription_info:
+        #     for k in d.keys():
```
sidler-mozilla: delete commented out code

src/shared/cfg.py
```diff
@@ -262,7 +262,7 @@ def AWS_EXECUTION_ENV(self):
 
     @property
     def SWAGGER_UI(self):
-        return self.DEPLOYED_ENV in ("stage", "qa", "dev")
+        return self.DEPLOYED_ENV in ("stage", "qa", "dev", "fab")
```
sidler-mozilla: good catch

src/hub/vendor/customer.py
```diff
@@ -46,56 +46,38 @@ def run(self) -> None:
...
-                products[user_plan["product"]] = product
-
-            plan_nickname = format_plan_nickname(
-                product_name=product["name"], plan_interval=user_plan["interval"]
-            )
-            formatted_nicknames.append(plan_nickname)
-
-        nicknames = " ".join(formatted_nicknames)
+            plan_amount = plan_amount + user_plan["plan_amount"]
+            nicknames = " ".join(user_plan["nickname"])
```
jackiemunroe: Is there a reason we went back to using Plan.nickname here? This particular Stripe value will be meaningless to FxA and Salesforce since all of their other interactions with SubHub use the formatted nickname.

src/shared/cfg.py
marty331: Thanks.

src/hub/vendor/customer.py
marty331: Updated @sidler-mozilla

src/hub/vendor/customer.py
marty331: Yes there is!  It's already written the correct way in src/sub/payments.py ln 199 and that is written to the deletedUsers table, so this is just concatenating the correct nickname values.

PR #323, opened 2019-10-19
Fix up containers

marty331:
- Add command to run sub as doit localsub
- Add command to run hub as doit localhub
- Update docs to reflect the change


src/sub/app.py
```diff
@@ -186,3 +186,4 @@ def after_request(response):
     app.debug = True
     app.use_reloader = True
     app.run(host="0.0.0.0", port=CFG.LOCAL_FLASK_PORT)
+    logger.info("sub running", app=app)
```
sidler-mozilla: don't want the same thing in hub/app.py?

src/hub/app.py
```diff
@@ -21,6 +21,7 @@
 
 logger = get_logger()
 
+
```
sidler-mozilla: just cause?

dodo.py
```diff
@@ -8,7 +8,9 @@
 import itertools
 import contextlib
 import threading
+import atexit
```
sidler-mozilla: are you using this?

marty331: Nope, forgot to remove.

src/sub/app.py
marty331: Yes, changing.

src/hub/app.py
marty331: Not sure what you are calling out?

Issue #305, opened 2019-10-10
The user is billed again when he resubscribes after canceling a subscription


madalincm:
#### Affected versions:
- Firefox Release 69.0.1
- Firefox Beta 70.0b9
- Firefox Nightly 71.0a1

#### Affected Platforms:
- Windows 10 (64-bit)
- MacOS 10.14.5
- Linux Ubuntu 16.04 (64-bit)

#### Precondition
User is logged in in the Guardian VPN homepage and he does have an active subscription.
https://stage.guardian.nonprod.cloudops.mozgcp.net/vpn

#### Steps to reproduce
1. Navigate to subscription page and cancel subscription for Firefox Guardian
2. From the same page Resubscribe to Firefox Guardian

#### Expected result:
The user is not billed again so no Payment Received email is sent to the user. 

#### Actual result:
When the user resubscribes a payment received email is sent
![userBilledAgainAfterResubscribe](https://user-images.githubusercontent.com/15685441/66553118-35dd0180-eb53-11e9-9a22-228a57ba9b51.gif)

stlhood:
Being tracked in JIRA as COPS-778.

jackiemunroe:
Update: The customer is not billed again, but receives an email indicating a purchase/charge upon subscription being updated

madalincm:
> Update: The customer is not billed again, but receives an email indicating a purchase/charge upon subscription being updated
That is correct, I have checked the stripe dashboard and the user is not billed again.
We still consider this as must fix for the release as the user will receive false information regarding how he is billed so I'm adding the major label.

jackiemunroe:
This issue is currently blocked by designs for a resubscription email.
JIRA issue tracking implementation: COPS-786

stlhood:
Content for the resubscription email is linked in COPS-786. The design should follow our standard email template.

stlhood:
Removed "Major" tag because the customer is not actually billed twice.

This issue will be resolved in the current sprint, via COPS-786.

stlhood:
COPS-786 is complete and closed. The user now receives an email that explains that they have not been charged again.

SoftVision-LorandJanos:
This issue is no longer reproducible, using the latest Firefox Release (70.0.1). Verified on Windows 10 x64.

PR #330, opened 2019-10-25
feat hub/vendor/customer.py: added support for sending reactivation event

jackiemunroe:
Added support to send subscription reactivation event data to Salesforce/Basket
Refactored StripeCustomerSubscriptionUpdated class
Improved code coverage

https://jira.mozilla.com/browse/COPS-815
https://github.com/mozilla/subhub/issues/305
https://github.com/mozmeao/basket/issues/362


src/hub/vendor/customer.py
```diff
@@ -257,159 +256,224 @@ def run(self) -> None:
...
-                self.get_latest_invoice(invoice_id)
-            else:
+            latest_invoice = stripe.Invoice.retrieve(id=invoice_id)
+            logger.info("get latest invoice", latest_invoice=latest_invoice)
+
+            if latest_invoice.charge is not None and latest_invoice.number is not None:
                 return latest_invoice
+
+            attempts -= 1
+            time.sleep(0.2)
```
marty331: In get_latest_invoice() you are implementing exponential backoff, you should use shared/vendor - retrieve_stripe_invoice.

src/hub/vendor/customer.py
```diff
@@ -257,159 +256,224 @@ def run(self) -> None:
...
+                latest_charge.payment_method_details.card.last4 is not None
+                or latest_charge.payment_method_details.card.brand is not None
             ):
-                time.sleep(0.2)
-                self.get_latest_charge(charge)
-            else:
                 return latest_charge
+
+            attempts -= 1
+            time.sleep(0.2)
```
marty331: Same as above, should use shared/vendor - retrieve_stripe_charge ln 346

src/hub/vendor/customer.py
jackiemunroe: removed method and updated references to use shared/vendor method

src/hub/vendor/customer.py
jackiemunroe: removed method and updated references to use shared/vendor method

PR #338, opened 2019-10-30
Remove dynalite

bowlofstew:
This pull request (PR) is the initial work required to:

- Remove dynalite 
- Correct an issue where dynalite's port is held open in the sub-hub-shared test component split

This PR represents the creation of a new test module which is stored in `shared` and not `shared/tests` to call that out.  This was done as it is a shared test fixture versus shared test but this is debatable and easily moved.  Upon the completion of this work, the plan is to merge master into the sub-hub-shared split branch and resolve the socket not closed on the shared module with this work.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/338)

### DIFF
--- a/src/hub/verifications/events_check.py
+++ b/src/hub/verifications/events_check.py
@@ -1,7 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
+import sys
 import time
 import stripe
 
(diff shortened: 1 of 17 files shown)


src/hub/verifications/events_check.py
```diff
@@ -1,13 +1,14 @@
...
-
+import sys
 import time
-import stripe
-
 from abc import ABC
 from datetime import datetime, timedelta
 from typing import Dict, Any
+
+import stripe
```
sidler-mozilla: import block out of order

src/sub/app.py
```diff
@@ -87,9 +87,7 @@ def create_app(config=None) -> connexion.FlaskApp:
     # configure_logger()
     logger.info("creating flask app", config=config)
     region = "localhost"
-    host = f"http://localhost:{CFG.DYNALITE_PORT}"
-    if is_docker():
-        host = f"http://dynamodb:{CFG.DYNALITE_PORT}"
+    host = f"http://dynalite:{CFG.DYNALITE_PORT}" if is_docker() else CFG.DYNALITE_URL
```
sidler-mozilla: love the ternary here ;)

bowlofstew: Thanks!

PR #364, opened 2019-11-08
feature: update plan and subscription responses

jackiemunroe:
* added plan and product metadata to plan and subscription responses to support the implementation of tiered plans
* extracted methods for plan and subscription formatting to make verifying response formatting easier
* updated and added tests

https://jira.mozilla.com/browse/COPS-860

### DIFF
--- a/src/sub/payments.py
+++ b/src/sub/payments.py
@@ -9,6 +9,7 @@ import json
 from datetime import datetime
 from typing import List, Dict, Any, Optional
 from stripe import Customer, Product
+from stripe.util import convert_to_dict
 from flask import g
 
 from sub.shared.vendor import (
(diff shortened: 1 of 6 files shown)


src/sub/payments.py
```diff
@@ -393,52 +397,21 @@ def create_return_data(subscriptions) -> JsonDict:
...
-                        "plan_id": subscription["plan"]["id"],
-                        "status": subscription["status"],
-                        "subscription_id": subscription["id"],
-                        "cancel_at_period_end": subscription["cancel_at_period_end"],
-                        "failure_code": intents["failure_code"],
-                        "failure_message": intents["failure_message"],
-                    }
+                    format_subscription(subscription, product, intents)
                 )
                 continue
```
marty331: continue is not required and can have undesired effects.

src/sub/payments.py
```diff
@@ -537,33 +510,46 @@ def create_update_data(customer) -> Dict[str, Any]:
...
-                        "plan_id": subscription["plan"]["id"],
-                        "status": subscription["status"],
-                        "cancel_at_period_end": subscription["cancel_at_period_end"],
-                        "subscription_id": subscription["id"],
-                        "failure_code": intents["failure_code"],
-                        "failure_message": intents["failure_message"],
-                    }
+                    format_subscription(subscription, product, intents)
                 )
                 continue
```
marty331: same as above, continue is not required and can have unexpected effects.

src/sub/payments.py
jackiemunroe: updated

src/sub/payments.py
jackiemunroe: updated

PR #388, opened 2019-11-19
COPS-894

marty331:
Card type is not capitalized in receipt emails
- Get correct brand value to send to Salesforce
- Add tests to cover new code for format branding
- Fix tests that required adjustments for branding changes


src/shared/utils.py
```diff
@@ -25,3 +25,24 @@ def format_plan_nickname(product_name: str, plan_interval: str) -> str:
 
 def get_indempotency_key() -> str:
     return uuid.uuid4().hex
+
+
+def format_brand(brand: str) -> str:
```
bowlofstew: One thing is not like the other.  This seems to be in line with vendor utils, credit utils, etc and not so much like the `get_indempotency_key` above.  That is my one and only minor nitpick here.  This being said, I will approve it given the expediency to demo and we can figure it out post demo given it is only a movement of a function across files.

marty331: Moved to vendor_utils.py

src/shared/vendor_utils.py
```diff
@@ -0,0 +1,24 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+def format_brand(brand: str) -> str:
+    """
+    Format brand for emails prior to sending to Salesforce
+    :param brand:
+    :return:
+    """
+    brand_list = [
```
sidler-mozilla: Stripe doesn't provide a function call to handle this mapping?

PR #427, opened 2019-12-16
COPS-852 Update sub/hub app.py tests

marty331:
- Updated tests for sub/app.py
- Updated tests for hub/app.py
- Added new method to handle GetError when no db connection exists

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -19,6 +19,12 @@ exclude_lines =
     if .debug:
     raise NotImplementedError
     if __name__ == .__main__.:
+    logger.
+    stripe.log
+    stripe.verify_ssl_certs
+    stripe.api_base
+    from
+    import
 omit =
     */tests/*
     */.venv/*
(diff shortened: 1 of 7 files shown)


.coveragerc
```diff
@@ -19,6 +19,12 @@ exclude_lines =
     if .debug:
     raise NotImplementedError
     if __name__ == .__main__.:
+    logger.
```
bowlofstew: Nice updates!

src/hub/app.py
```diff
@@ -82,6 +89,7 @@ def create_app(config=None) -> Any:
     region = "localhost"
     host = f"http://dynamodb:{CFG.DYNALITE_PORT}" if is_docker() else CFG.DYNALITE_URL
     stripe.api_key = CFG.STRIPE_API_KEY
+    logger.debug("aws", aws=CFG.AWS_EXECUTION_ENV)
```
bowlofstew: This isn't directly related to the test work but I dig it.

PR #429, opened 2019-12-18
Database test coverage

bowlofstew:
This pull request (PR) strives to provide over 0 test coverage which it previously had.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/429)

### DIFF
--- a/src/shared/tests/conftest.py
+++ b/src/shared/tests/conftest.py
@@ -1,0 +1,30 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+import signal
+import subprocess
+import uuid
+import logging
+import json
+import psutil
+import pytest
+import stripe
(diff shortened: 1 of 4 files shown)


src/shared/tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+THIS_PATH = os.path.join(os.path.realpath(os.path.dirname(__file__)))
+
+
+def pytest_configure():
+    os.environ["AWS_ACCESS_KEY_ID"] = "fake"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "fake"
+    os.environ["USER_TABLE"] = "users-testing"
+    os.environ["DELETED_USER_TABLE"] = "deleted-users-testing"
+    os.environ["ALLOWED_ORIGIN_SYSTEMS"] = "Test_system,Test_System,Test_System1"
```
marty331: Shouldn't we make ln29-33 CFG values?

bowlofstew: Good point, these were copy and pasted to pull in the fixture registration.  I will update now as those aren't needed.

PR #435, opened 2020-01-06
Initial Terraform boostrap and DynamoDB setup

bowlofstew:
This pull request (PR) provides:

- The initial Terraform bootstrapping
- Creation of the DynamoDB tables from Terraform
- Removal of the DynamoDB tables from Serverless

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/435)


terraform/global/s3/main.tf
```diff
@@ -0,0 +1,26 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+resource "random_id" "terraform_state_bucket_name" {
+  prefix      = "${var.deployed_env}-terraform-"
+  byte_length = 8
+}
+
+resource "aws_s3_bucket" "terraform_state" {
+  bucket = random_id.terraform_state_bucket_name.hex
```
limed: Instead of `random_id.terraform_state_bucket_name.hex` we usually do something like `app-state-${account_id}` thats the convention we have been using for buckets names for state

bowlofstew: Cool, I will update and follow that convention.

terraform/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+  type        = "string"
+  default     = "us-west-2"
+  description = "the region where to provision the stack."
+}
+
+######################################
+# DynamoDB Configuration
+######################################
+
+variable "USERS_TABLE_READ_CAPACITY" {
```
limed: Does your variable need to be caps?

terraform/global/s3/main.tf
```diff
@@ -0,0 +1,21 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+resource "aws_s3_bucket" "terraform_state" {
+  bucket = "app-state-${account_id}"
```
limed: Sorry I should be more explicit here but instead it should look like this

```
data "aws_caller_identity" "current" {}

resource "aws_s3_bucket" "terraform_state" {
  bucket = "subhub-state-${data.aws_caller_identity.current.account_id}"
}
```

Something like that

terraform/main.tf
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+# https://www.terraform.io/docs/providers/aws/
+provider "aws" {
+  alias  = "configured"
```
limed: You don't really need this unless you are going to reference the "configured" part somewhere

terraform/dynamodb.tf
```diff
@@ -0,0 +1,47 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+resource "aws_dynamodb_table" "users_table" {
```
limed: So i guess my question is why is there a `dyanmodb.tf` file and then there is something similar in `terraform/global/dynamodb/main.tf`

What are the differences?

terraform/main.tf
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+
+# https://www.terraform.io/docs/providers/aws/
+provider "aws" {
+  region = "us-west-2"
```
limed: We generally like to set  this as 

```
provider "aws" {
   region = var.region
}
```

And then have a `variables.tf` file and set a default value to the region variable to `us-west-2` this is still ok though

terraform/dynamodb.tf
bowlofstew: The global dynamo is the Terraform locks table for deployments.  This table represents the constructs an application level, users object table for consumption by the application.

PR #436, opened 2020-01-08
Fix broken Pipeline tests

marty331:


docker-compose.yml
```diff
@@ -97,3 +97,9 @@ services:
     container_name: stripe
     ports:
       - "$STRIPE_MOCK_PORT:12112"
+
+  wiremock:
```
bowlofstew: Can be nuked now, correct?

src/shared/cfg.py
```diff
@@ -224,9 +224,7 @@ def DYNALITE_PORT(self):
 
     @property
     def SALESFORCE_BASKET_URI(self):
-        return self(
-            "SALESFORCE_BASKET_URI", "https://subhub.free.beeceptor.com?api-key="
-        )
+        return self("SALESFORCE_BASKET_URI", "http://127.0.0.1:3000/posts?api-key=")
```
bowlofstew: Needed still?

src/shared/dynamodb.py
```diff
@@ -17,7 +17,7 @@
 logger = get_logger()
 
 
-def random_label(length):
+def random_label(length: int) -> str:
```
bowlofstew: nice improvement, courtesy of mypy.

docker-compose.yml
marty331: Yes, good catch, fixing now.

src/shared/cfg.py
marty331: Nope, will fix.

src/shared/dynamodb.py
marty331: yes, so thankful for mypy and it's courteous behavior...

PR #471, opened 2020-04-15
bug(invoice email) - ensure invoice emails are sent for subscriptions of any interval

jackiemunroe:
- Remove check to only send emails for subscriptions that are billed on a "day" or "month" based interval

Because:
- We now want to support a wider variety of subscription billing intervals for our products being offered

Fixes: https://github.com/mozilla/fxa/issues/4940

### DIFF
--- a/src/hub/vendor/invoices.py
+++ b/src/hub/vendor/invoices.py
@@ -78,75 +78,36 @@ class StripeInvoicePaymentSucceeded(AbstractStripeHubEvent):
         Parse the event data sent by Stripe contained within self.payload
         :return True to indicate successfully sent
         """
+        invoice = self.payload.data.object
         logger.debug("invoice payment succeeded", payload=self.payload)
-        subscription_id = self.payload.data.object.subscription
+        subscription_id = invoice.subscription
         subscription = retrieve_stripe_subscription(subscription_id)
         if subscription:
             plan = subscription.get("plan")
-            created = self.payload.data.object.created
             customer = subscription.get("customer")
             user_id = None
             metadata = customer.get("metadata")
             if metadata:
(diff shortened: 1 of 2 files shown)


src/hub/vendor/invoices.py
```diff
@@ -78,75 +78,36 @@ def run(self) -> bool:
...
-                routes = [StaticRoutes.SALESFORCE_ROUTE]
-                self.send_to_routes(routes, json.dumps(data))
-                return True
-            else:
-                logger.error(
-                    "plan type not supported", payment_type=is_subscription_new
-                )
-        return False
 
-    def determine_new(self, subscription: Subscription, created_date: int) -> str:
```
jackiemunroe, line 122: This was removed because it was no longer being used.

src/hub/vendor/invoices.py
```diff
@@ -78,75 +78,36 @@ def run(self) -> bool:
...
         subscription = retrieve_stripe_subscription(subscription_id)
         if subscription:
             plan = subscription.get("plan")
-            created = self.payload.data.object.created
             customer = subscription.get("customer")
             user_id = None
             metadata = customer.get("metadata")
             if metadata:
                 user_id = metadata.get("userid")
-            is_subscription_new = self.determine_new(subscription, created)
```
jackiemunroe, line 92: The invoice contains the reason for why it was created. If 'billing_reason' is 'subscription_create', then it is a new subscription. Otherwise it is a recurring subscription. The date comparisons that were being done in the `determine_new` method are not necessary to determine if it is a new subscription. This logic is already being used elsewhere in this file to determine if it is a new subscription.

PR #476, opened 2020-04-23
Feature/initial cd

bowlofstew:
This pull request (PR) adds support for release based branches for CD.  This will make it easier for other developers not familiar with the project to get up and running, reduce overall maintenance costs, etc per OKR #2.

This change is [Reviewable](https://reviewable.io/reviews/mozilla/subhub/476)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,33 +2,38 @@ language: python
 dist: xenial
 sudo: required
 python:
- - "3.7"
+- "3.7"
 services:
- - docker
+- docker
+notifications:
+  email:
+    on_success: never
+    on_failure: change
 addons:
   apt:
-    sources: ['ubuntu-toolchain-r-test']
-    packages: ['bash', 'coreutils', 'awscli']
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - bash
+    - coreutils
+    - awscli
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


bin/deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "No DEPLOY_ENV to set."
+    ;;
+esac
+
+if [ -z "$DEPLOY_ENV" ]; then
+      echo "Not deployinng"
+else
+      echo "Deploying to $DEPLOY_ENV"
+      doit deploy
+fi
```
The-smooth-operator, line 28: Missing new line

.travis.yml
```diff
@@ -2,33 +2,38 @@ language: python
 dist: xenial
 sudo: required
 python:
- - "3.7"
+- '3.7'
```
The-smooth-operator, line 5: Just out of curiosity, I see here you change `"` per `'`. But some of the new introduced parameters above are using `"`. Is this intended?

bin/deploy.sh
bowlofstew: On it, thanks!

.travis.yml
bowlofstew: Good catch.  I will go over and make it overall consistent now.

